PR #42, opened 2026-03-22
feat: add configurable DNS provider with safe apply flow and auto-restart

Himanth-reddy:
## Why

I was facing network-related issues in the app, so I added DNS provider selection to improve reliability and make it easier to switch resolvers when connectivity is unstable or region-dependent.

## What changed

1. Added configurable DNS provider support across app networking.
2. Added DNS-over-HTTPS support and resolver options (System DNS, Cloudflare, Google, AdGuard).
3. Wired DNS selection into Settings UI.
4. Added a confirmation step before applying DNS changes.
5. Automatically restarts the app after DNS change so all clients use the new resolver cleanly.
6. Updated image/network client wiring to use the selected DNS consistently.
7. Moved DNS switch/connection pool eviction work off the main thread to avoid UI-thread network exceptions.

## User impact

1. Users can change DNS from Settings.
2. DNS changes are applied consistently after confirmation and restart.
3. Better resilience for users with ISP DNS issues, blocked domains, or unstable network resolution.

## Motivation / context

This was implemented specifically to address network problems I was experiencing, where changing DNS can help recover API/image connectivity and stop request failures.

## Testing done

1. Built app variant successfully after changes.
2. Verified DNS selection flow from Settings.
3. Verified apply confirmation and automatic restart flow.

## Notes

No functional changes outside DNS/network resolver behavior and the related Settings flow.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -226,6 +226,7 @@ dependencies {
     implementation("com.squareup.retrofit2:retrofit:2.9.0")
     implementation("com.squareup.retrofit2:converter-gson:2.9.0")
     implementation("com.squareup.okhttp3:okhttp:4.12.0")
+    implementation("com.squareup.okhttp3:okhttp-dnsoverhttps:4.12.0")
     implementation("com.squareup.okhttp3:logging-interceptor:4.12.0")
 
     // Coroutines
--- a/app/src/main/kotlin/com/arflix/tv/ArflixApplication.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ArflixApplication.kt
@@ -1,7 +1,6 @@
 package com.arflix.tv
 
 import android.app.Application
-import android.graphics.Bitmap
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.work.Configuration
 import androidx.work.Constraints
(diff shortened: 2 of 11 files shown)


ProdigyV21:
Thanks for the PR - configurable DNS is a useful feature. However there are a few issues that need fixing before merging:

API key leak: apiDnsLoggingInterceptor logs full URLs (including API keys) in release builds. Needs BuildConfig.DEBUG gate.
Cache crashes: client creates a new OkHttpClient + Cache on every access. Multiple Cache instances on the same directory will cause file-lock crashes. Should be a singleton.
Split Coil loaders: The Hilt-injected ImageLoader and Coil.setImageLoader() create two separate instances with split caches - prefetches won't warm AsyncImage's cache.
ANR risk: runBlocking in Application.onCreate() can freeze startup if DataStore I/O is slow.
Restart unnecessary: Since OkHttpProvider.dns is already a dynamic delegator, evicting the connection pool should suffice without killing the process.

Happy to help revise if you want to address these.

Himanth-reddy:
About the restart, I have checked if it dynamically reloads but it is not working as intended. So I just moved to restarting the app. I will revise this according to other changes and update the pull request according to the latest version 1.9.3

Himanth-reddy:
@ProdigyV21 I’ve addressed all the requested changes and pushed the updates to this branch.
The PR is now up to date with your feedback, including the DNS/settings adjustments and merge-conflict resolution with main.
Please take another look when you have a moment, and let me know if you want any further tweaks.

PR #48, opened 2026-03-24
Fix/search portrait cards

EierKopZA:
Title:

Fix: Portrait cards clipping in Search UI

Description:

Issue: When the CardLayoutMode was set to Portrait/Poster, the taller cards and their two lines of text were getting clipped off at the bottom of the screen inside the RowsLayer container.

Fix: > * Increased the rowHeight calculation for portrait mode to 340.dp to accommodate the taller images and multi-line titles.

Added extra bottom contentPadding (24.dp) to the LazyRow so the cards have room to expand during the focus scale animation without hitting the bounds of the layout.


EierKopZA:
image

ProdigyV21:
Has been added in 1.9.4. Thanks!

PR #49, opened 2026-03-24
Add Preferred Max Quality setting and refactor Auto-Play logic

EierKopZA:
1. New Feature: Preferred Max Quality Setting
Settings UI: Added a new toggle in Settings > General (Index 10) below the Min Quality setting. Users can cycle through Any, 720p, 1080p, and 4K.

State Management: Integrated the setting into SettingsViewModel and DetailsViewModel via DataStore for persistent user preference.

Cloud Sync: Added the new preference to CloudSyncRepository to ensure quality limits sync across devices.

2. Refactored Auto-Play Engine (DetailsScreen.kt)
The Issue: The previous logic used broad string matching (e.g., .contains("4K")), which caused "fake 4K" links—like those labeled DS4K (Downsampled 4K)—to be misidentified as 2160p and blocked when a 1080p limit was set.

The Fix: * Implemented word-boundary Regex (\b) in the scoring logic to accurately identify resolutions and ignore false positives.

Updated the LaunchedEffect to use a strict range filter: score in minThreshold..maxThreshold.

The app now ignores any stream exceeding the user's "ceiling" and automatically selects the highest available stream within the allowed range.

Developer Notes
While the Auto-Play logic is now robust and correctly identifies/filters streams based on the filename, there is still a visual inconsistency in the UI badges within StreamSelector.kt and SourceInfoOverlay.kt. Some 1080p files are still rendering with an orange "4K" badge due to the provider's metadata. The engine ignores these for auto-play correctly, but a final Regex pass on the UI components would be beneficial to align the badges with the now-fixed selection logic.

How to test
Go to Settings > General and set Preferred Max Quality to 1080p.

Open a title with multiple resolutions (e.g., Zootopia 2).

Verify that auto-play successfully bypasses 2160p/4K links and starts a 1080p link instead.

Set the limit to 720p and verify it downshifts the selection further.


EierKopZA:
Hey! While testing the other features, I noticed a couple of navigation bugs and pushed two quick QoL (Quality of Life) fixes to this branch:

Watchlist Focus Trap: Refined the grid bounds using onPreviewKeyEvent so the D-pad no longer gets permanently stuck inside the grid when trying to navigate Left or Up to the sidebar/top bar.

Search Navigation: Swapped remember to rememberSaveable for the Search focusZone so the app remembers you were in the results list when pressing back from the Details screen, instead of resetting to the keyboard.

ProdigyV21:
Am still looking into this, will see to it for 1.9.5

PR #79, opened 2026-03-30
Fix general settings scroll: DNS Provider focus

EierKopZA:
## Problem\nDNS Provider was unreachable because Skip Profile Selection and DNS rows shared index 12 and general section maxIndex capped at 12.\n\n## Solution\n- Make DNS Provider focused index 13\n- Update content focus bounds for general section from 12 to 13 in both scroll and key handling.\n\n## Validation\n- :app:compilePlayDebugKotlin succeeded.


Himanth-reddy:
Looks like this issue has already been resolved in v1.9.6 (DNS Provider is now focusable and clickable via D-pad).

@ProdigyV21 could you confirm if this PR can be closed?

EierKopZA:
Yeah, @ProdigyV21 fixed it right after I made the PR haha. I will close this.

ProdigyV21:
Yes, just when I was done with the fix I saw the PR xD

PR #84, opened 2026-03-30
Add translation to ARVIO and bugFix on language for loading content when loading profile

Eragorn:
I know it's a big pull request
I love this app but i'm French and i miss my good old language

I make all the static field i found (there are maybe some that i missed) translatable with android ressources

i also make a fix for language content settings which wasn't fetch on loading profile only when going into settings view

THANKS A LOT FOR YOUR APP MEN


Eragorn:
Preview : 
image

Himanth-reddy:
Thanks for the big localization pass here. The PR is definitely moving ARVIO in the right direction, and I verified that it builds successfully on my side for both debug flavors.

Before merge, I think there are a few behavior issues we should tighten up:

The “Content Language” setting currently also changes the global app interface language. In the UI this setting reads like it should only affect titles/descriptions/metadata, so this will be surprising for users and also means one profile can change the app language for every profile.

Preinstalled catalog titles are now localized, but the catalog merge logic preserves old stored titles as if they were user renames. That means existing profiles may keep old English labels instead of updating to the selected language.

The profile-switch language fix still looks a bit race-prone. Home navigation can happen as soon as the active profile changes, but mediaRepository.contentLanguage is updated in a separate coroutine, so the first load may still use the previous profile’s language.

I’d be happy to approve after a follow-up pass on those points, because the translation work itself is valuable and the PR is otherwise in a good direction.

Himanth-reddy:
please try to split changes like this into smaller PRs when possible. A PR with translation changes plus behavior changes plus settings/profile logic is much harder to review and test safely. Smaller focused PRs make it much easier for us to review and merge faster.

Eragorn:
> Thanks for the big localization pass here. The PR is definitely moving ARVIO in the right direction, and I verified that it builds successfully on my side for both debug flavors.
> 
> Before merge, I think there are a few behavior issues we should tighten up:
> 
> The “Content Language” setting currently also changes the global app interface language. In the UI this setting reads like it should only affect titles/descriptions/metadata, so this will be surprising for users and also means one profile can change the app language for every profile.
> 
> Preinstalled catalog titles are now localized, but the catalog merge logic preserves old stored titles as if they were user renames. That means existing profiles may keep old English labels instead of updating to the selected language.
> 
> The profile-switch language fix still looks a bit race-prone. Home navigation can happen as soon as the active profile changes, but mediaRepository.contentLanguage is updated in a separate coroutine, so the first load may still use the previous profile’s language.
> 
> I’d be happy to approve after a follow-up pass on those points, because the translation work itself is valuable and the PR is otherwise in a good direction.

Ok i will look into it when i have the time

Eragorn:
> please try to split changes like this into smaller PRs when possible. A PR with translation changes plus behavior changes plus settings/profile logic is much harder to review and test safely. Smaller focused PRs make it much easier for us to review and merge faster.

Yeah i know how hard it is to review 
i'm very sorry 🥺

I was focused on translation in 1.9.2 when the content wasn't localizable

Then i rebase and got myself a bit overwhelmed with all the conflicts

Eragorn:
Hello @Himanth-reddy, i made change following your comments : 
- Language change is now a full parameter
- Language name choices is translatable
- The language changement is activated directly after param change
- add a runBlocking when changing mediaRepo language to avoid concurency

ProdigyV21:
First — thank you for the huge localization effort. A proper French translation is real work and genuinely improves ARVIO for non-English users. The previous reviewer confirmed the PR builds cleanly for both flavors, so the translation pass itself is clearly usable.

**That said, this PR still cannot be safely merged in its current form** for a few concrete reasons:

**1. Scope is too large to review as a single unit (42 files, +2186 / -751)**
The diff mixes three very different kinds of change:
- ~1200 lines of pure localization (`values/strings.xml`, `values-fr/strings.xml`, `stringResource` replacements across every screen)
- Behavior changes to how Content Language works (it's now tied to the app UI language, not just metadata)
- Profile/startup logic changes (`runBlocking` on the profile-switch path, catalog merge behavior)

A regression in any of (2) or (3) would affect every user, not just French speakers. **Please split this into at least two PRs:**
- **PR A:** pure translation / `stringResource` extraction + `values-fr/strings.xml`. This is low-risk and should be mergeable fast.
- **PR B:** the Content Language / profile-switch / catalog-merge behavior changes. This is where the risk lives and it deserves a focused review.

**2. `Content Language` now also changes the full app UI language**
In the UI this setting reads like it should only affect titles/descriptions/metadata from TMDB, but in this PR it also swaps the app interface locale. That is a surprising behavior change for existing users, and because it's a profile-scoped setting it means one profile switching its content language will change the UI language for every profile that uses the app afterward. Either rename the setting (e.g. `App Language`) or keep interface language as a separate device-level preference from content/metadata language.

**3. Preinstalled catalog titles get stuck with old English labels after upgrade**
Preinstalled catalog titles are now localized via resources, but the catalog-merge logic preserves existing stored titles as if they were user renames. That means existing profiles will keep their old English labels instead of picking up the translated ones. The merge logic needs to distinguish between a default title (which should be replaced by the current-locale resource) and a genuinely user-renamed title (which should be preserved).

**4. `runBlocking` on the profile-switch path is a yellow flag**
Calling `runBlocking` on the main/UI path to avoid the `mediaRepository.contentLanguage` race is effectively hiding a concurrency bug rather than fixing it. On slow devices or with a cold DataStore this will cause visible UI jank on profile switch. The correct fix is to `await` the language update before triggering home navigation (e.g. make the profile-switch flow `suspend` and sequence the two operations explicitly in a single coroutine), not to block the caller thread.

**5. CI has not been run against the current head of this branch** (`statusCheckRollup` is empty). Please push a fresh commit to re-trigger the Release Gate workflow so we can confirm both `playDebug` and `sideloadDebug` still build green against current `main`.

Once (1) is split and (2)–(4) are addressed, the translation half (PR A) should be very fast to approve and merge. Thanks again for the work — we do want this in the app.

PR #107, opened 2026-04-02
feat: sync app clock format with device settings

EierKopZA:
This PR updates the top bar clock to respect the user's Android device time setting, automatically switching between 12-hour and 24-hour formats accordingly.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
+}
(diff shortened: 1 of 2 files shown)


Himanth-reddy:
Great improvement. Syncing the top bar clock with the device 12/24-hour preference is the right behavior for Android TV and improves UX consistency.

One small suggestion: instead of manually switching between HH:mm and h:mm a with SimpleDateFormat, consider using DateFormat.getTimeFormat(context) for fully locale-aware formatting (including regional AM/PM conventions).

Also, if the VS Code settings file was added unintentionally, it may be worth removing it from this PR to keep the diff focused.

Himanth-reddy (approved):

PR #108, opened 2026-04-02
feat: Add Subtitle Offset Setting to Android TV App

EierKopZA:
### Description
This pull request adds a new quality-of-life feature allowing users to vertically offset their subtitles on the screen during playback. 

Some users prefer subtitles sitting lower out of the active content edge, or slightly higher to match external borders. To accommodate this, a 5-step configurable "Subtitle Offset" setting has been added alongside existing options (Subtitle Color, Subtitle Size).

### Changes Made:
- Added `subtitleOffset` DataStore preference mapping in `SettingsViewModel` and `PlayerViewModel`.
- Embedded `SettingsRow` within `GeneralSettings` section inside the `SettingsScreen` UI layer.
- Subtitle views are configured with `setBottomPaddingFraction` within `PlayerScreen.kt`, dynamically reacting to the offset fractions (`0.02f` to `0.18f`) requested. 

No existing settings routines or ExoPlayer dependencies were modified or broken.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt
@@ -1628,47 +1628,46 @@ fun PlayerScreen(
                         useController = false
                         setKeepContentOnPlayerReset(true)
                         resizeMode = playerResizeMode
-
-                        // Enable subtitle view with Netflix-style: bold white text with black outline
-                        subtitleView?.apply {
-                            // Read subtitle appearance from user settings
-                            val subSizeSp = when (subtitleSizePref) {
-                                "Small" -> 18f; "Large" -> 30f; "Extra Large" -> 36f; else -> 24f
-                            }
-                            val subFgColor = when (subtitleColorPref) {
-                                "Yellow" -> android.graphics.Color.YELLOW
-                                "Green" -> android.graphics.Color.GREEN
-                                "Cyan" -> android.graphics.Color.CYAN
-                                else -> android.graphics.Color.WHITE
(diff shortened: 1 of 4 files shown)


Himanth-reddy:
Great feature addition. The subtitle offset setting is a useful quality-of-life improvement and fits naturally with the existing subtitle controls.

One small technical note: ensure the subtitle offset is applied in both the player view initialization and the update path, so changes take effect immediately during playback without needing to recreate the player view.

ProdigyV21:
Thanks for this PR — the subtitle offset feature itself looks good and CI was green. However, after merging #120 (UI Mode Warning Dialog), this branch now conflicts with `main` in `SettingsScreen.kt`. It's now in a `CONFLICTING / DIRTY` state and can't be merged as-is.

**To unblock merge, please:**

1. **Rebase on latest `main`** and resolve the conflicts in `app/src/main/kotlin/com/arflix/tv/ui/screens/settings/SettingsScreen.kt`. The conflicts will be in the General settings section where #120 added UI mode confirmation handling.
2. After rebasing, **double-check the focus indices** in `GeneralSettings`. Inserting your Subtitle Offset row shifts every item after it down by one, so any `focusedIndex == N` checks and the max-index navigation clamp (look for a comment like `0 -> N // General: N items`) need to be updated to keep DNS Provider and every row below the new one reachable via D-pad. This has been a recurring issue in other recent PRs that added settings rows (#110, #112), so worth verifying explicitly.
3. *(Optional, from the earlier review)* Apply the subtitle offset in the PlayerView **update path** in addition to initialization, so changes take effect live during playback without recreating the view.

Once rebased with correct focus indices, this should be ready to merge.

PR #109, opened 2026-04-02
feat: add regex-based stream autoplay filtering (mirrored from NuvioTV)

EierKopZA:
This PR implements a new "Auto-Play Match Mode" that allows for advanced stream filtering using regex patterns during autoplay. This provides users with much finer control than simple quality thresholds (inspired by NuvioTV's implementation).

Key Features:

Regex Presets: Added a selection of common filter presets in Settings (e.g., Any 1080p+, 4K / Remux, No CAM/TS, HEVC / x265).
Device-Specific Persistence: Settings are stored locally via DataStore (not synced to cloud) to ensure that specific playback preferences stay on the intended device.
Advanced Matching: Automatically builds a searchable metadata blob for every stream (Addon + Quality + Source + URL) and matches it against the user's regex pattern.
Smart Fallback: If a regex is active but no matching streams are found, the app intelligently falls back to the manual source picker.
Files Changed:

SettingsViewModel.kt: Added local persistence logic.
SettingsScreen.kt: Integrated the "Match Mode" UI row and picker modal.
DetailsViewModel.kt: Exposed the regex state to the details view.
DetailsScreen.kt: Implemented the core regex comparison and auto-selection logic.


ProdigyV21:
The regex autoplay feature itself is a nice addition and the core logic in `DetailsScreen.kt` / `DetailsViewModel.kt` / `SettingsViewModel.kt` looks reasonable (device-scoped DataStore key, sensible fallback to the stream selector when no match, good preset list). However, **this PR cannot be merged in its current state** because it accidentally ships build/editor artifacts:

**Blocking — please remove these files from the PR:**

1. `.vscode/settings.json` — local editor config, shouldn't be in the repo.
2. `clean_out.txt` — committed as a **binary** file (appears to be gradle output).
3. `tmp_gradle_out.txt` — committed as a **binary** file (appears to be gradle output).

These look like accidental additions from your local gradle runs. You can drop them with:

\\\ash
git rm .vscode/settings.json clean_out.txt tmp_gradle_out.txt
git commit --amend --no-edit   # or a new commit
git push --force-with-lease
\\\

And consider adding them to `.gitignore` so it doesn't happen again.

**Also verify (same recurring gotcha as #110 / #112):** inserting the `Auto-Play Match Mode` row at `focusedIndex == 8` shifts every item after it down. Please confirm that the max-index clamp in `GeneralSettings` navigation is updated so the rows below remain reachable via D-pad, and that no other row still hard-codes `focusedIndex == 8`.

Once the three stray files are removed and the focus/clamp check is done, this should be straightforward to merge.

PR #110, opened 2026-04-02
feat: add toggle for trailer banner audio

EierKopZA:
This PR introduces a new user preference that allows users to toggle the audio for trailers playing in the Home Screen Hero Banner.

By default, trailers will remain muted to ensure a non-intrusive experience, but users can now choose to enable sound via the settings menu.

Key Changes:

New Preference: Added trailer_audio_enabled to the local DataStore. This setting is profile-specific.
Settings UI: Added a new "Trailer Audio" toggle in the General Settings > Playback section.
Audio Control: Updated the TrailerPlayer component to support an isMuted parameter. When unmuted, the trailer plays at a comfortable 50% volume (0.5f).
State Propagation: Updated HomeViewModel to load the current audio preference and pass it through to the TrailerPlayer on the Home screen.
Files Modified:

SettingsViewModel.kt: Added logic to manage and persist the new setting.
SettingsScreen.kt: Added the toggle UI row and updated the GeneralSettings layout.
HomeViewModel.kt: Updated to load the setting for the home screen UI state.
HomeScreen.kt: Passes the user's preference down to the banner player.
TrailerPlayer.kt: Added support for muting/unmuting the background video.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
@@ -57,8 +57,8 @@ class CloudSyncRepository @Inject constructor(
         val frameRateMatchingMode: String = "Off",
         val autoPlayNext: Boolean = true,
         val autoPlaySingleSource: Boolean = true,
-        val autoPlayMinQuality: String = "Any",
         val trailerAutoPlay: Boolean = false,
+        val trailerAudioEnabled: Boolean = false,
         val includeSpecials: Boolean = false,
         val iptvHiddenGroups: String = "",
         val iptvGroupOrder: String = ""
@@ -70,6 +70,8 @@ class CloudSyncRepository @Inject constructor(
         profileManager.profileStringKeyFor(profileId, "content_language")
     private fun trailerAutoPlayKeyFor(profileId: String) =
         profileManager.profileBooleanKeyFor(profileId, "trailer_auto_play")
+    private fun trailerAudioEnabledKeyFor(profileId: String) =
+        profileManager.profileBooleanKeyFor(profileId, "trailer_audio_enabled")
 
     private fun subtitleSizeKeyFor(profileId: String) =
         profileManager.profileStringKeyFor(profileId, "subtitle_size")
@@ -151,6 +153,7 @@ class CloudSyncRepository @Inject constructor(
                         contentLanguage = prefs[contentLanguageKeyFor(profile.id)] ?: "en-US",
 
                         trailerAutoPlay = prefs[trailerAutoPlayKeyFor(profile.id)] ?: false,
+                        trailerAudioEnabled = prefs[trailerAudioEnabledKeyFor(profile.id)] ?: false,
                         subtitleSize = prefs[subtitleSizeKeyFor(profile.id)] ?: "Medium",
                         subtitleColor = prefs[subtitleColorKeyFor(profile.id)] ?: "White",
                         iptvHiddenGroups = prefs[iptvHiddenGroupsKeyFor(profile.id)] ?: "",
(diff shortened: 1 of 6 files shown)


Himanth-reddy (changes requested):
Great feature! The trailer audio toggle logic is clean and well-integrated. Two quick fixes needed before merge:

## 1. Focus Index Collision (Critical)
The "Trailer Audio" toggle at `focusedIndex == 9` collides with "Match Frame Rate" (also `== 9`), making the audio toggle unreachable via D-pad. 
- Update "Match Frame Rate" to `focusedIndex == 10`
- Increment all subsequent items: Card Layout → 11, UI Mode → 12, Skip Profile Selection → 13, DNS Provider → 14
- Update max index comment: `0 -> 13 // General: 14 items` → `0 -> 14 // General: 15 items` (two locations in SettingsScreen.kt)

## 2. Trailer Aspect Ratio
Change [TrailerPlayer.kt](app/src/main/kotlin/com/arflix/tv/ui/components/TrailerPlayer.kt) `resizeMode` from `RESIZE_MODE_ZOOM` to `RESIZE_MODE_FILL` so the trailer crops to fill the entire TV screen without letterboxing.

## Summary
Everything else looks solid—audio state propagation is correct, DataStore persistence is safe, and the default (muted) behavior is user-friendly.

EierKopZA:
@Himanth-reddy done

Himanth-reddy:
Thanks for the update. I took another pass through the changes and wanted to leave a full review here instead of the shorter earlier note.

There was some confusion on my side around the trailer sizing behavior. So there is no issue there and no resize-mode change is needed in this PR. So Change the resize mode back to zoom

The remaining issues I’m seeing are:

## 1. General settings navigation still stops too early
The General section now includes the new `Trailer Audio` row and shifts the later items down, but the D-pad navigation limit still stops before the final `DNS Provider` row. As a result, `DNS Provider` is not reachable through normal remote navigation.

## 2. Trailer audio preference is not scoped to the active profile
The new setting is stored as a profile-scoped preference, but the Home screen loads it by checking whether *any* saved preference key ends with `_trailer_audio_enabled`. That means if one profile enables trailer audio, another profile can also end up reading it as enabled. The value should be loaded for the active profile only.

## 3. Trailer audio is not included in cloud sync / restore
`setTrailerAudioEnabled()` triggers cloud sync, but the new setting is not part of the cloud profile snapshot or restore path. So it looks persisted from the UI, but it will not survive sync/restore between devices the way the surrounding playback settings do.

## Scope
I’m also noticing that this branch contains additional settings-related changes beyond the trailer audio feature itself. Since the core trailer-audio feature is small and the extra settings work makes the diff harder to review, I’d recommend keeping this PR focused on the trailer audio toggle and moving the unrelated settings changes into a new pull request.

But I still checked the trailer cropping, it still does not fill the entire screen. If you can, make a new pull request with this update.

Once the navigation limit, active-profile loading, and sync coverage are fixed, this should be in much better shape.

ProdigyV21:
Thanks for this feature — the trailer audio toggle logic itself is clean and the default-muted behavior is a good choice. **This PR is still in `CHANGES_REQUESTED` state** from @Himanth-reddy's review, and the blocking issues haven't been resolved yet. Summarizing what's still needed before merge:

**1. D-pad navigation limit still stops too early**
The General section now includes the new `Trailer Audio` row and shifts the later items down, but the D-pad navigation limit still stops before the final `DNS Provider` row. As a result, `DNS Provider` is not reachable through normal remote navigation. Please update the max-index clamp in `GeneralSettings` to match the new row count. (This is the same recurring issue as #112 / #109 — any new settings row needs both the focus index remap *and* the clamp bump.)

**2. Trailer audio preference is not correctly scoped to the active profile**
The setting is stored as a profile-scoped preference, but `HomeViewModel` loads it by checking whether *any* saved preference key ends in `_trailer_audio_enabled`. That means if one profile enables trailer audio, every other profile will also read it as enabled. Please load the value for the active profile only (use the standard `profileManager.profileBooleanKey(...)` path used by the surrounding settings).

**3. Not included in cloud sync / restore**
`setTrailerAudioEnabled()` triggers cloud sync, but the new setting is not part of the cloud profile snapshot or the restore path. So it looks persisted from the UI, but it won't survive sync/restore between devices the way the surrounding playback settings do. Please add it to `CloudSyncRepository` snapshot and restore alongside the other playback prefs.

**4. Scope creep**
This branch contains additional settings-related changes beyond the trailer audio feature itself. To make review easier and reduce the risk surface of each merge, please either:
- move the unrelated settings changes into a separate PR, or
- explicitly list them in the description so reviewers know what extra ground to cover.

The trailer cropping / `resizeMode` point from the earlier review has been clarified — no change needed there in this PR (per Himanth's follow-up comment on 2026-04-03).

Once 1–4 are addressed, this is ready to merge.

PR #112, opened 2026-04-02
feat: implement episode spoiler protection

EierKopZA:
Adds a 'Hide Spoilers' toggle in Settings to blur/blackout unwatched episode thumbnails.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/details/DetailsScreen.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/details/DetailsScreen.kt
@@ -583,6 +583,7 @@ fun DetailsScreen(
                     reviews = uiState.reviews,
                     similar = uiState.similar,
                     similarLogoUrls = uiState.similarLogoUrls,
+                    hideEpisodeSpoilers = uiState.hideEpisodeSpoilers,
                     focusedSection = focusedSection,
                     buttonIndex = buttonIndex,
                     episodeIndex = episodeIndex,
@@ -921,6 +922,7 @@ private fun DetailsContent(
     reviews: List<Review>,
     similar: List<MediaItem>,
     similarLogoUrls: Map<String, String>,
+    hideEpisodeSpoilers: Boolean = false,
     focusedSection: FocusSection,
     buttonIndex: Int,
     episodeIndex: Int,
@@ -1198,7 +1200,7 @@ private fun DetailsContent(
                             EpisodeCard(
                                 episode = episode,
                                 isFocused = false,
-                                hideEpisodeSpoilers = uiState.hideEpisodeSpoilers,
+                                hideEpisodeSpoilers = hideEpisodeSpoilers,
                                 onClick = { onEpisodeClick(index) }
                             )
                         }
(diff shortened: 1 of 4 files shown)


Himanth-reddy:
Thanks for the PR. I tested the settings flow and also reviewed the implementation, and there are still a few issues that should be fixed before merge.

## 1. `Hide Spoilers` is visible but does not actually toggle
The new row is shown in Settings and is focused at index `9`, but the OK/Enter handler for General settings was not updated to match the new item order. Pressing select on that row still triggers the old action for `Match Frame Rate` instead of toggling `Hide Spoilers`.

## 2. `DNS Provider` is not reachable
The new item ordering shifted `DNS Provider` down, but the D-pad navigation limit for the General section still stops before that row. So users cannot navigate to `DNS Provider` anymore.

## 3. Spoiler protection does not fully hide spoiler content
In `EpisodeCard`, the blackout overlay is added over the thumbnail, but the episode metadata is still rendered afterward. That means the episode title, air date, rating, and overview can still appear on top of the spoiler overlay for unwatched episodes.

## 4. The new setting is not included in cloud sync / restore
`hide_episode_spoilers` is saved locally and triggers sync, but it is not included in the cloud snapshot / restore path. So it will not persist across sync/restore the same way the surrounding settings do.

Please fix the Settings navigation/action mapping, fully hide episode metadata when spoiler protection is enabled, and add the new preference to the cloud sync flow.

ProdigyV21:
Thanks for this PR — episode spoiler protection is a nice feature and CI is green. **It cannot be merged yet**, though, because of the issues @Himanth-reddy already raised on 2026-04-03. Posting them again so they're consolidated and easy to address:

**1. `Hide Spoilers` toggle does nothing when selected**
The new row is displayed in Settings at `focusedIndex == 9`, but the OK/Enter handler for the General section was not updated to match the new item order. Pressing select on that row still triggers the old action for `Match Frame Rate` instead of toggling `Hide Spoilers`. Every subsequent `focusedIndex == N` branch also needs to be incremented, and the max-index clamp comment (`0 -> N // General: N items`) needs to be bumped.

**2. `DNS Provider` becomes unreachable**
The new item ordering shifts `DNS Provider` down, but the D-pad navigation limit for the General section still stops before that row. As a result users can no longer navigate down to DNS Provider at all. The navigation max index needs to be raised to match the new row count.

**3. Spoiler protection does not fully hide spoiler content**
In `EpisodeCard`, the blackout overlay is added over the thumbnail, but the episode metadata (title, air date, rating, overview) is still rendered afterward on top of the overlay. For unwatched episodes the title/overview should also be hidden or replaced with a placeholder when spoilers are enabled — otherwise the thumbnail blur is cosmetic.

**4. New preference isn't cloud-synced**
`hide_episode_spoilers` is saved locally and triggers cloud sync, but it's not included in the cloud snapshot / restore path. So toggling it looks like it persists from the UI, but it won't survive sync/restore between devices the way the surrounding settings do. Please add it to the cloud snapshot and restore flow like the other playback preferences.

Once those four points are addressed this should be in good shape to merge.

PR #166, opened 2026-05-08
fix: restore details page layout and remove dates from poster card subtitles

EierKopZA:
**Description:**

Fixes the details page content rows being positioned too high on Android TV by restoring vertical spacing and reducing the vertical footprint of poster cards in similar/collection rows.

**Changes:**
- Restored `contentRowBottomPadding` from `0.dp` to `12.dp` (combined with upstream's `contentRowHeight` reduction to `260-330dp`)
- Removed year and media type labels from `SimilarMediaCard` subtitles — cards now show only the title under the poster image, saving ~20dp per card and allowing content rows to sit lower without clipping


ProdigyV21:
Has already been adjusted in latest commits. Check again if my fixes were enough.

PR #170, opened 2026-05-09
Ai subtitles + upgrade Exo to 1.9.0

silentbil:
AI Subtitle Translation

-  Adds real-time AI-powered subtitle translation using Groq or Google Gemini directly inside the player — no external apps, no manual copy-paste.
- Upgrade media3Version to version 1.9.0

  ✅ No more missing subtitles                                                                                                                                                                                                     
  ✅ No more subtitle timing issues

  How it works:
  - Enable AI Translation in  Settings → Integrations → AI Subtitles
  - Set your free Groq API key or paid Gemini key (or scan a QR code from your phone)
  - Play any content with built-in subtitles — translation happens live as cues appear

  Features:
  - Auto-select mode: automatically activates AI translation on playback when enabled                                                                                                                                           
  - Strip hearing-impaired annotations from translated text
  - Subtle AI badge while translating, toast notification on first success or error
  - Key entry via on-screen keyboard or phone QR scan
  - Falls back to addon subtitles when no built-in tracks are available
  - Off by default — zero impact for users who don't opt in

selection:

<img width="886" height="528" alt="image" src="https://github.com/user-attachments/assets/c5513241-0503-430e-a256-dbbd65b697cf" />

settings:

<img width="1822" height="955" alt="image" src="https://github.com/user-attachments/assets/21c17f61-b329-4d73-a578-b22ad021eda3" />

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -28,8 +28,7 @@ android {
         applicationId = "com.arvio.tv"
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         // Fire TV devices can be as low as Android 7.1 (API 25) or lower depending on model/OS.
-        // Lower minSdk to maximize compatibility and avoid "There was a problem parsing the package".
-        minSdk = 21
+        minSdk = 23
         targetSdk = 35
         versionCode = 260
         versionName = "1.9.91"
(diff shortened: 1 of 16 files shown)


ProdigyV21:
Great feature, would be great to have.

The two remaining things I’d like clarified or fixed before merge:

1. `minSdk` changed from 21 to 23. That means ARVIO would stop installing on Android 5.0/5.1 devices and require Android 6.0+. If Media3 1.9.0 or the new AI subtitle code truly requires API 23, that is fine, but please confirm. If not required, please keep `minSdk = 21` so older TV boxes do not lose support.

2. AI translation errors seem to be swallowed. `SubtitleTranslationManager` falls back to the original subtitle text when the API call fails, but it does not call the failure callback. So invalid API keys, rate limits, or network/API failures may not show the user any error toast. Please call `onBatchResult(false, error)` on failed translation so users know why AI translation is not working.

silentbil:
@ProdigyV21 

1. minSdk 23 is required — media3-datasource-okhttp:1.9.0 declares it in its own manifest
2. good catch, fixed

PR #175, opened 2026-05-11
Apply focus border colour override to all focusable components

EierKopZA:
All components that draw their own `.border()` with hardcoded `Color.White` or 
`Pink` (which is actually `AccentWhite`) now call `resolveFocusBorderColor()`, 
which reads `LocalFocusBorderColorOverride` — the same CompositionLocal that the 
existing `arvioFocusable()` modifier already respects.

### Components covered
- **SettingsScreen**: SettingsRow, SettingsToggleRow, AddonRow, AccountActionRow, 
  SettingsActionRow, AccountRow, SettingsChip, CatalogDiscovery* (3 variants), 
  modal text fields, paste button, regex editor fields, option picker
- **SearchScreen**: GlowChip (filter chips)
- **PersonModal**: HorizontalKnownForCard

Backward compatible — falls back to the original colour when no override is set.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/components/PersonModal.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/components/PersonModal.kt
@@ -47,6 +47,7 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.input.key.Key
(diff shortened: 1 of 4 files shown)


EierKopZA:
@copilot apply changes based on the comments in [this thread](https://github.com/ProdigyV21/ARVIO/pull/175#pullrequestreview-4264632498)

EierKopZA:

PR #176, opened 2026-05-11
Fix ProfileSelectionScreen to respect OLED black background setting

EierKopZA:
Replaces the hardcoded BackgroundDark gradient background in 
ProfileSelectionScreen with appBackgroundDark(), which reads the 
LocalOledBlackBackground CompositionLocal.

When the user enables "OLED black background" in Settings, this screen
now renders pure black (Color.Black) like every other screen already does.


EierKopZA:

PR #178, opened 2026-05-11
feat: add 'Your Watch History' home row with 'Because You Watched' recommendations grid

EierKopZA:
### Summary
Adds a **"Your Watch History"** row to the home screen (positioned after Continue Watching) showing all items from the user's watch history. Clicking any item opens a **recommendations grid** powered by TMDB's `getRecommendations()` API — the same engine used by the existing "More Like This" in the Details screen.

### Changes

| File | Change |
|------|--------|
| `ui/screens/recommendations/RecommendationsScreen.kt` | **New** — ViewModel + TvLazyVerticalGrid backed by `MediaRepository.getSimilar()` |
| `navigation/AppNavigation.kt` | Added `Screen.Recommendations` route + NavHost composable |
| `ui/screens/home/HomeViewModel.kt` | Injects "Your Watch History" row from cached Continue Watching items |
| `ui/screens/home/HomeScreen.kt` | Wired D-pad + click routing for watch_history category items |

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/navigation/AppNavigation.kt
+++ b/app/src/main/kotlin/com/arflix/tv/navigation/AppNavigation.kt
@@ -25,6 +25,7 @@ import com.arflix.tv.ui.screens.home.HomeScreen
 import com.arflix.tv.ui.screens.login.LoginScreen
 import com.arflix.tv.ui.screens.player.PlayerScreen
 import com.arflix.tv.ui.screens.collections.CollectionDetailsScreen
+import com.arflix.tv.ui.screens.recommendations.RecommendationsScreen
 import com.arflix.tv.ui.screens.search.SearchScreen
 import com.arflix.tv.ui.screens.settings.SettingsScreen
 import com.arflix.tv.ui.screens.tv.live.LiveTvScreen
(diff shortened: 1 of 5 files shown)


EierKopZA:
@copilot apply changes based on the comments in [this thread](https://github.com/ProdigyV21/ARVIO/pull/178#pullrequestreview-4265539497)

EierKopZA:
@ProdigyV21 - I couldn't test because no catalogs were loading in the debug build, Not sure if its a TMDB key issue.

This should theoritically work. Worth adding (can revert if it doesn't do what it's supposed to).

ProdigyV21:
Nice feature idea, and it does compile now. I think it needs a few fixes before merge though.

Main issue: the new “Your Watch History” row is currently built from `lastContinueWatchingItems`, so it is not really full watch history. It mostly duplicates Continue Watching. If the goal is “Because You Watched”, it should use actual watch history / watched items, not only the active Continue Watching list.

Second issue: mobile/touch clicks probably won’t always open the recommendations screen. The click handler checks `focusState.currentRowIndex` to decide if the item came from the `watch_history` row, but mobile rows don’t really update that TV focus state. Better to pass the category id into the click callback, or handle it inside `MobileHomeRowsLayer` using the local `category`.

Small routing issue too: `mediaTitle` is manually decoded with `URLDecoder.decode()` after Navigation already gives a decoded argument. Titles with `%` or `+` can break/display wrong. Better to use `Uri.encode()` when creating the route and then read the argument directly.

Build checks passed:

So it is close, but I would fix those before merge.

EierKopZA:
@copilot apply changes based on the comments in [this thread](https://github.com/ProdigyV21/ARVIO/pull/178#pullrequestreview-4307818937)

ProdigyV21:
This is much closer now. The watch history row is now based on real watch history instead of just Continue Watching, and the mobile/touch routing looks improved too.

Still not merge-ready yet though: GitHub reports this PR as conflicted, and a real merge into current `main` conflicts in `HomeViewModel.kt`. Please rebase onto current main and resolve that first.

After that, please also clean up:
- `RecommendationsScreen` handles Back twice (`BackHandler` + `Key.Back`), which can double-pop.
- API/network failures are still shown as “No recommendations found”, so errors should have a separate state.
- New user-facing strings should use string resources.

Feature idea is good, but it needs the rebase + these small fixes before merge.

EierKopZA:
@ProdigyV21 do I need to update anything here?

PR #182, opened 2026-05-12
fix: persist skip profile selection state

EierKopZA:
Summary: trigger cloud sync when Skip Profile Selection is toggled so cloud restore does not overwrite the new value on next startup. Testing: ./gradlew test fails in this environment with message 25.0.2 before and after this change.


EierKopZA:

PR #184, opened 2026-05-12
fix: keep app update install progress visible

EierKopZA:
Keep the app update modal visible after Install, persist installer status (installing/pending/success/failed), and surface that status in Settings so installation progress does not disappear. In this environment, gradlew test fails early with 25.0.2 before task execution.


EierKopZA:
@copilot apply changes based on the comments in [this thread](https://github.com/ProdigyV21/ARVIO/pull/184#pullrequestreview-4270494477)

EierKopZA:
@ProdigyV21 - I cannot test this since I don't get updates in my debug build. It compiles and merges without any issues so probably worth a shot. We can revert if it doesn't do what it's supposed to.

Issue #180, opened 2026-05-12
Some Requests


reddit-reaper:
Hello,

I wanted to reach out and say thanks for all the work you're putting into ARVIO. It’s a great app and I’m really enjoying using it. I had a few requests for ARVIO, if possible, for the future that I think would make the experience even better:

- **Smoother tile navigation**: Right now, moving through tiles feels a bit "hard." Making the animations a bit more fluid would make browsing feel a lot smoother.
- **Current Tile selection**: It would be cool if the tile you’re currently focused on became a bit bigger than the others so it really stands out.
- **Anime Subtitles (libass)**: If you could add libass support for ExoPlayer/Media3, it would help a ton with rendering stylized anime subtitles correctly.
- **Alternative Player**: Adding libvlc as an optional player would be a great backup for different file types. External player has already been suggested so i wont add to that but +1 it
- **Internal Watchlist**: It would be awesome to have a separate watchlist or library feature that doesn't rely on Trakt (kind of like how Stremio handles its library).
- **Separate Categorization/Catalogs**: If it’s possible to separate Anime (and K-Dramas if possible) into their own sections instead of having them mixed in with regular movies and shows, that would be a huge help for organization.
- **AniList Syncing**: Adding support to sync with AniList (similar to Trakt) would be a massive quality-of-life update for those of us who track our anime there.
- **Shuffle Playback:** Some people like to binge random episodes of shows :)

Thanks for considering these! Excited to see how the app keeps growing. Keep up the awesome work!

PR #192, opened 2026-05-14
feat: Add stylized subtitles toggle

Himanth-reddy:
Addresses stylized anime subtitles request from #180

## Description
Adds a new setting to enable stylized subtitles for ASS/SSA subtitle formats.

## Changes
* Added `subtitle_stylized` string resources.
* Added `subtitleStylized` preference to `PlayerUiState` and `SettingsUiState`.
* Updated `PlayerScreen` to conditionally apply `setApplyEmbeddedStyles(true)` and `setApplyEmbeddedFontSizes(true)` on `SubtitleView` when enabled. When disabled, the fallback custom sizes and colors via `CaptionStyleCompat` are used.
* Updated `SettingsScreen` to include the Stylized Subtitles toggle in both the TV layout and Mobile layout, properly managing D-pad navigation indexes for the TV layout.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
@@ -106,6 +106,7 @@ class CloudSyncRepository @Inject constructor(
         val subtitleColor: String = "White",
         val subtitleStyle: String = "Bold",
         val subtitleOffset: String = "Low",
+        val subtitleStylized: Boolean = true,
         val cardLayoutMode: String = CARD_LAYOUT_MODE_LANDSCAPE,
         val frameRateMatchingMode: String = "Off",
         val autoPlayNext: Boolean = true,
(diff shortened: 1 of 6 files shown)


ProdigyV21:
Nice feature, especially for anime ASS/SSA subtitles. The merge build passes, but I think two things should be fixed before merging.

1. Plain subtitles can ignore the existing user subtitle appearance settings  
When `subtitleStylizedPref` is true, `PlayerScreen` only enables embedded styles/font sizes. It does not apply the user’s normal subtitle size/color/style as the fallback style. Since this setting defaults to true, normal SRT/VTT subtitles may stop respecting the existing subtitle appearance settings. The comment says user prefs are used as fallback for plain SRT/VTT, but the code only applies them in the non-stylized branch.

Please apply the user `CaptionStyleCompat`/fixed text size first, then enable embedded styles/font sizes only when stylized mode is on.

2. New setting does not cloud sync  
`toggleSubtitleStylized()` calls `syncLocalStateToCloud()`, but `CloudSyncRepository.CloudProfileSettings` does not include/export/import `subtitleStylized`. So the setting is local-only and will not restore on another device.

Please add `subtitleStylized` to the cloud profile settings model and read/write `subtitle_stylized` during export/import.

Checks passed:
- `git diff --check`

Himanth-reddy:
Addressed both review items in Himanth-reddy:feat/stylized-subtitles.

- Plain SRT/VTT fallback styling now applies the user subtitle CaptionStyleCompat and fixed text size before toggling embedded ASS/SSA styles, so stylized mode can still honor user appearance settings for plain subtitle formats.
- Added subtitleStylized to CloudProfileSettings and wired subtitle_stylized through cloud export/import so the setting restores across devices.

Checks passed locally:
- git diff --check
- ./gradlew :app:compileSideloadDebugKotlin
- ./gradlew :app:compilePlayDebugKotlin

PR #215, opened 2026-05-17
feat: add custom User-Agent setting

chillpill244:
## Summary

- Adds a **Custom User-Agent** setting in the Network section of Settings (both TV and mobile layouts) that lets users override the HTTP User-Agent string sent with all requests
- Stores the value in DataStore (`custom_user_agent_global` key) and restores it on startup before any network calls
- Injects the configured value via an OkHttp interceptor on both the app client and the playback client — derived clients (IPTV, Xtream, stream resolution) inherit it automatically without a client rebuild

## Changes

**`OkHttpProvider`** — `DEFAULT_USER_AGENT` + `USER_AGENT_PREF_KEY` constants; `@Volatile _customUserAgent` with getter/setter; `userAgentInterceptor` as the first interceptor in `buildAppClient()` and `buildPlaybackClient()`

**`ArflixApplication`** — loads and applies the saved UA from DataStore at startup alongside DNS initialisation

**Repositories / Player** — replaced all hardcoded UA strings in `CatalogRepository`, `IptvRepository`, `MediaRepository`, `StreamRepository`, and `PlayerScreen` with `OkHttpProvider.userAgent`

**`SettingsViewModel`** — `customUserAgent` in `SettingsUiState`; loads, persists, and applies via `setCustomUserAgent()`

**`SettingsScreen`** — Custom User-Agent row in the Network section (TV D-pad + mobile); `CustomUserAgentDialog` text-input dialog; callbacks threaded through `MobileSettingsLayout` → `MobileSettingsSubPage`

**`strings.xml`** — `custom_user_agent` and `custom_user_agent_desc` resources

## Test plan

- [x] Settings → Network on TV: row present and D-pad navigable
- [x] Settings → Playback & Controls on mobile: row present under CONTROLS
- [x] Enter a custom value, restart — value persists
- [x] Verify UA appears in stream/IPTV requests (logcat or proxy)
- [x] Clear the field — app falls back to the default Chrome UA
- [x] No regression on DNS provider or other Network rows

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ArflixApplication.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ArflixApplication.kt
@@ -82,13 +82,17 @@ class ArflixApplication : Application(), Configuration.Provider, ImageLoaderFact
         // a single volatile assignment.
         OkHttpProvider.init(this)
 
-        // Initialize global DNS provider from DataStore before network calls.
+        // Initialize global DNS provider and user agent from DataStore before network calls.
         appScope.launch(Dispatchers.IO) {
             val prefs = settingsDataStore.data.first()
             val dnsKey = androidx.datastore.preferences.core.stringPreferencesKey(OkHttpProvider.DNS_PROVIDER_PREF_KEY)
             val dnsPref = prefs[dnsKey]
             val provider = OkHttpProvider.parseDnsProvider(dnsPref)
             OkHttpProvider.setDnsProvider(provider)
+
+            val uaKey = androidx.datastore.preferences.core.stringPreferencesKey(OkHttpProvider.USER_AGENT_PREF_KEY)
+            val savedUserAgent = prefs[uaKey]?.takeIf { it.isNotBlank() } ?: OkHttpProvider.DEFAULT_USER_AGENT
+            OkHttpProvider.setCustomUserAgent(savedUserAgent)
 
             runCatching { OkHttpProvider.dns.lookup("image.tmdb.org") }
         }
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CatalogRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CatalogRepository.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
+import com.arflix.tv.network.OkHttpProvider
 import okhttp3.OkHttpClient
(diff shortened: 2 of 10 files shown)


ProdigyV21:
It is a nice feature, but am not sure people really need this? I never encountered such issue before. Could you explain how and why this could help people?

chillpill244:
> It is a nice feature, but am not sure people really need this? I never encountered such issue before. Could you explain how and why this could help people?

This is a feature that I also observed in many IPTV apps. Many IPTV providers send the 200 response based on the User Agent in HTTP headers. The one that I'm using does that. The video playback only works if we set the user agent is set to `Smarterspro`.

Guiro28:
I also need this feature for one of my IPTV subscriptions. My provider blocks playback based on the user agent.

Collier14:
Yes. I do not need this but all IPTV apps carry this as some providers require this for the playlist to work. 
We 100 percent need this on arvio.

ProdigyV21:
Alright, can merge this then. 

Only thing I would change before merge: please keep the old Chrome-style User-Agent as the default/fallback instead of changing the default to `Mozilla/5.0 (Android TV; ARVIO)`.

Some stream/IPTV hosts are picky and work better when the request looks like Chrome. The custom setting can still allow users to enter whatever they want, but the default should stay the most compatible one so existing streams do not regress.

chillpill244:
> Alright, can merge this then.
> 
> Only thing I would change before merge: please keep the old Chrome-style User-Agent as the default/fallback instead of changing the default to `Mozilla/5.0 (Android TV; ARVIO)`.
> 
> Some stream/IPTV hosts are picky and work better when the request looks like Chrome. The custom setting can still allow users to enter whatever they want, but the default should stay the most compatible one so existing streams do not regress.

Updated

ProdigyV21:
The default User-Agent issue is fixed now. It uses the old Chrome-style UA as fallback, so that part is good.

But this PR still needs a rebase because latest main now conflicts with it. SettingsScreen.kt uses index 34 for Trailer Delay, and this PR also uses 34 for Custom User-Agent. Please rebase onto latest main and move Custom User-Agent to a free focus index.

Small safety improvement: the interceptor should only set the global User-Agent if the request does not already have one, so provider-specific stream headers do not get overwritten.

ProdigyV21:
Merged a hardened replacement in #240.

I did not merge this branch directly because it was no longer mergeable against latest `main`, reused row id `34` which now belongs to Trailer Delay, and changed default User-Agent behavior in ways that could break IPTV/source-specific headers.

#240 keeps the custom User-Agent setting, but leaves the value blank by default, preserves existing IPTV/M3U/EPG defaults when blank, and avoids overwriting request-specific User-Agent headers.

PR #218, opened 2026-05-18
Improve non-Trakt cross-device cloud sync responsiveness

EierKopZA:
## Summary
Improves non-Trakt cloud sync responsiveness for Continue Watching progress across devices (e.g., TV to phone resume).

## Changes
- Reduce watch-history save interval during active playback from 15s to 5s.
- Add immediate persistence when playback position changes significantly (seek jump >= 20s).
- Reduce realtime watch_history event debounce from 5s to 1s.
- Tighten local self-echo suppression window from 3s to 1.5s to avoid delaying legitimate remote updates.
- Update HomeViewModel inline comment to reflect the faster write cadence.

## Why
Previously, stacked throttles caused noticeably stale resume positions between devices for non-Trakt users. These changes shrink the stale window and make cross-device resume behavior feel much closer to instant.

## Scope
This PR is intentionally scoped to non-Trakt watch-progress sync paths and realtime propagation timing.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/RealtimeSyncManager.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/RealtimeSyncManager.kt
@@ -62,7 +62,8 @@ class RealtimeSyncManager @Inject constructor(
         // meaningfully increasing load (one GET every 45s while signed in).
         private const val PERIODIC_SYNC_INTERVAL_MS = 45_000L
         private const val DEBOUNCE_MS = 2_000L
-        private const val WATCH_HISTORY_DEBOUNCE_MS = 5_000L
+        private const val WATCH_HISTORY_DEBOUNCE_MS = 1_000L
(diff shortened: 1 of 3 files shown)


ProdigyV21:
Thanks, the idea is good and faster cross-device resume would definitely improve the app.

I’m a bit worried about the write volume though. Changing active playback saves from every 15s to every 5s means each active viewer goes from around 4 watch-history writes/minute to around 12 writes/minute, plus extra writes after seek jumps.

For a small user base that is fine, but with thousands of users this can become a lot of Supabase writes/realtime events very quickly. I think 5s is probably too aggressive for production.

Can we make it a safer middle ground?

Suggested approach:
- Use 10s for normal active playback saves instead of 5s.
- Still save immediately on pause/stop.
- Still save immediately after a big seek jump.
- Keep the realtime debounce improvement, or use 2s if needed.

That should still make TV-to-phone resume feel much faster than before, without tripling the normal write load.

EierKopZA:
Hey @ProdigyV21 - happy for you to implement with the 10s suggestion.

PR #219, opened 2026-05-18
Fix non-Trakt Continue Watching movie visibility

EierKopZA:
## Summary
Fixes missing movie entries in Continue Watching for non-Trakt users.

## Changes
- Keep non-Trakt CW items when `resumePositionSeconds > 0`, even if `progress` is missing or zero.
- Derive a minimal in-progress value when only position metadata exists.
- Stop dropping entries when `title` is blank by using a safe fallback title.

## Why
Some in-progress movies were never shown on Home because filtering relied only on progress percentage and mapping dropped entries with incomplete metadata.

## Scope
Only non-Trakt Continue Watching movie visibility behavior.


EierKopZA:
@copilot apply changes based on the comments in [this thread](https://github.com/ProdigyV21/ARVIO/pull/219#pullrequestreview-4308014771)

PR #222, opened 2026-05-18
perf: debounce syncAddonCatalogs() with addon list fingerprint

EierKopZA:
## perf/debounce-sync-catalogs

Debounce `syncAddonCatalogs()` with addon list fingerprint.

`syncAddonCatalogs()` iterates every addon manifest, flatMaps `manifest.catalogs`, constructs `CatalogConfig` entries, reads hidden-addon preferences from DataStore, diffs against persisted catalogs, and optionally writes back to DataStore — all of which is wasted work when the installed addon list is identical between calls.

Added a `@Volatile` `lastSyncedAddonFingerprint` field. On entry, a joinToString fingerprint of addon IDs is compared; if unchanged since last sync, returns `false` immediately.

**Impact**: Sub-millisecond exit path for redundant calls. `syncAddonCatalogs()` is called on every `loadHomeData()` invocation; when the addon setup is stable (which is the common case), the entire ~50-line processing pipeline is bypassed.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CatalogRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CatalogRepository.kt
@@ -53,6 +53,12 @@ class CatalogRepository @Inject constructor(
         MediaRepository.buildPreinstalledDefaults().associateBy { it.id }
     }
 
+    // Debounce guard: skip syncAddonCatalogs() when addon list hasn't changed.
+    // syncAddonCatalogs() iterates every addon manifest, flatMaps catalogs, and
+    // diffs against persisted state — all of which is wasted work when the
+    // installed addon list is identical between calls.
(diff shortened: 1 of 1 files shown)


ProdigyV21:
The debounce fingerprint only uses addon IDs:

`addons.joinToString(",") { it.id }`

But `syncAddonCatalogs()` depends on more than IDs: active profile, addon enabled/installed state, addon URL, addon type, and manifest catalog contents. If the same addon IDs remain but a user disables an addon, switches profile, or the manifest/catalog list changes, this can skip sync and leave stale/missing catalogs.

Example: profile A syncs catalogs, then profile B loads home with the same addon IDs. Because `lastSyncedAddonFingerprint` is global, profile B can skip sync before its own catalog list is populated.

Please include profile id + relevant addon fields/catalog manifest data in the fingerprint, or clear/reset this debounce when profile/catalog/addon state changes.

PR #228, opened 2026-05-19
fix(iptv): eliminate 5–30 s source-picker latency and thread-safety bugs

chillpill244:
## Summary

- **Persistent movie source cache** (LRU-256, 24 h TTL, profile-scoped `SharedPreferences`) — repeat movie source clicks return in ~5 ms instead of 5–30 s
- **VOD ID index** built at catalog load — reduces TMDB/IMDb movie lookups from O(n) linear scan over 60 k items to O(1)
- **Series episode fast path** (`tryFastResolveEpisodeFromCache`) — reads stored bindings + cached episode lists without network; cuts cached episode resolution from 500 ms–1 s to <20 ms
- **Multi-binding** for series — stores all matched catalog variants (4K, dub, split-season entries) per show; every variant appears as a separate picker row after first resolve
- **Settings → Refresh IPTV** now calls `purgeAllIptvSourceCaches()` to hard-reset all IPTV caches before re-fetching

Thread-safety and correctness fixes (from code review):
- Fix TOCTOU race on `iptvMovieSourcePersistJob` cancel/assign — add `iptvPersistLock`
- Fix `clearAll()` cross-profile prefs wipe — selective key deletion instead of `prefs.edit().clear()`
- Fix `writeSeriesBinding` silently discarding bindings past cap-of-8 — merge existing IDs before `take(maxSeriesBindingsPerKey)`
- Fix `ensureVodCatalogIndex` size comparison → reference equality (`===`)
- Fix `readSeriesBinding` prefs read outside `synchronized` block (write-after-read race)
- Cache `profileIdHash()` result per profile ID — avoids repeated `MessageDigest` allocation
- Move `gson.fromJson` outside `synchronized` in `hydrateIptvMovieSourceCache`
- Replace `profileManager.activeProfileId.first()` with `getProfileIdSync()` (4 sites)
- Replace `System.err.println` with `Log.d(TAG, ...)` (70 sites)
- Fix `ClassCastException` crash when provider returns `"info": []` in `get_series_info` response — use `get("info")?.takeIf { it.isJsonObject }?.asJsonObject` instead of unsafe `getAsJsonObject()`

## Files changed

- `IptvRepository.kt` — all cache, index, resolver, and fix changes
- `SettingsViewModel.kt` — wires `purgeAllIptvSourceCaches()` into the force-refresh path

## Test plan

- [x] Cold start → open a movie → click Sources → IPTV results appear in ~5 ms on second visit
- [x] Open a TV series → click several episodes → each resolves in <20 ms after the first
- [x] Series with multiple catalog variants (4K / dubbed) shows multiple picker rows
- [x] Switch profiles → verify IPTV cache isolation (Profile A results don't bleed into Profile B)
- [x] Settings → IPTV → Refresh → all IPTV caches purge and re-fetch from provider
- [x] Provider with `"info": []` on episode objects → no crash, episode resolves normally

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/IptvRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/IptvRepository.kt
@@ -34,6 +34,8 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.launch
+import com.arflix.tv.network.OkHttpProvider
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.xmlpull.v1.XmlPullParser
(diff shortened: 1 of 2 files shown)


ProdigyV21:
Hey there, thanks for the pull. This could be a great one.

This looks like a strong IPTV performance improvement. It merges cleanly and both Play/Sideload Kotlin compile tasks pass.

One thing I would fix before merging:

The persistent movie source cache key does not include `year` when there is no TMDB/IMDb id. That means title-only matches can collide for remakes or movies with the same name, for example `Dune 1984` vs `Dune 2021`. If the first one is cached, the second one could get the wrong IPTV source.

Please include the movie year in `iptvMovieSourceCacheKey()` for title-only cache entries. After that, I think this is good to merge and should noticeably improve IPTV source loading speed.

chillpill244:
> Hey there, thanks for the pull. This could be a great one.
> 
> This looks like a strong IPTV performance improvement. It merges cleanly and both Play/Sideload Kotlin compile tasks pass.
> 
> One thing I would fix before merging:
> 
> The persistent movie source cache key does not include `year` when there is no TMDB/IMDb id. That means title-only matches can collide for remakes or movies with the same name, for example `Dune 1984` vs `Dune 2021`. If the first one is cached, the second one could get the wrong IPTV source.
> 
> Please include the movie year in `iptvMovieSourceCacheKey()` for title-only cache entries. After that, I think this is good to merge and should noticeably improve IPTV source loading speed.

Updated to also include the year from titles in cache key

ProdigyV21:
Merged this via replacement PR #235 with the final cache-key fix included.

The extra fix passes the real `year` argument into the title-only IPTV movie source cache key, falling back to parsing the title only when needed. This avoids collisions for remakes with the same title.

Verification passed:
- `./gradlew.bat :app:compilePlayDebugKotlin :app:compileSideloadDebugKotlin`
- `git diff --check`

Thanks for the performance work.

PR #229, opened 2026-05-19
fix: restore IPTV navigation, catchup sync, and mobile interactivity

Himanth-reddy:
This PR addresses multiple core issues with IPTV playback, UI navigation, and catchup synchronization across both Android TV and mobile layouts:

- **Unified LazyColumn EPG**: Refactored the `LiveTvScreen` to merge the channel list and EPG grid into a single vertically scrolling LazyColumn, drastically improving scroll physics, preventing lag, and ensuring perfect sync between components.
- **D-pad Navigation Fixes**: Re-implemented `BackHandler` behaviors and focus requesters for the TV overlay to ensure zapping, fast-scrolling, and escaping the EPG grid returns smoothly to the channel column without getting stuck.
- **Timezone Offset Alignment**: Fixed a critical bug where Xtream Catch-up returned a `200 OK` but wouldn't play due to UTC timezone drift. The parser now dynamically computes the server's local timezone offset from EPG timestamps and formats `/timeshift/` URLs appropriately.
- **Mobile Touch Controls & Fullscreen HUD**: Engineered complete touch gesture support for mobile devices:
  - Ensured past archive/EPG programs are clickable on mobile screens.
  - Tapping the fullscreen layout now surfaces the HUD overlay.
  - Added a dedicated floating 'Back/Exit' action button in the HUD for mobile users to quickly exit full-screen mode.
- **Automatic XTREAM Resolution**: Catchup format correctly resolves to XTREAM when numeric stream IDs and standard URL paths are detected, even without standard M3U catchup flags.


ProdigyV21:
Merged this via replacement PR #230 with the two follow-up fixes included:

- Xtream server offset is now scoped per provider/credentials instead of one global key.
- Xtream streams with numeric stream IDs are treated as catchup-capable even when the playlist does not expose explicit catchup flags, so past programs can be loaded and selected consistently.

Thanks for the PR.

PR #231, opened 2026-05-19
Feature/catalog move to top bottom

EierKopZA:
### Summary
Adds two new action buttons to the catalog context menu in Settings — "Move to Top" (SkipPrevious icon) and "Move to Bottom" (SkipNext icon) — allowing users to jump a catalog directly to either end of the visible list without repeatedly pressing the existing single-step Up/Down buttons.

### Changes

**CatalogRepository.kt** (lines 825+)
- Added `moveCatalogToTop(catalogId: String): Boolean` — removes the catalog from its current position and inserts it at the position of the first visible catalog
- Added `moveCatalogToBottom(catalogId: String): Boolean` — removes the catalog and inserts it after the last visible catalog
- Both methods filter for visible catalogs (hidden catalogs are not reordered) and map visible indices back to full-list indices

**SettingsViewModel.kt** (lines 1692+)
- Added `moveCatalogToTop(catalogId: String)` and `moveCatalogToBottom(catalogId: String)` delegates that launch coroutines calling the repository methods with `syncLocalStateToCloud(silent = true)`

**SettingsScreen.kt** (multiple locations)
- Updated D-pad focus action indices from 5 to 7 actions per catalog row:
  - `0` = Edit, `1` = Move to Top, `2` = Move Up, `3` = Move Down, `4` = Move to Bottom, `5` = Layout, `6` = Delete
- Added `SkipPrevious` icon chip before ArrowUp and `SkipNext` icon chip after ArrowDown
- Updated all D-pad boundary checks (`catalogActionIndex < 4` → `< 6`)
- Updated `when (catalogActionIndex)` dispatch to handle new indices 1 and 4
- Updated both `CatalogsSettings` call sites with new `onMoveCatalogToTop`/`onMoveCatalogToBottom` callbacks

### Testing
- ✅ Kotlin compilation passes (`compileSideloadDebugKotlin` — `BUILD SUCCESSFUL in 6s`)
- D-pad focus flows predictably: Edit → **Top** → Up → Down → **Bottom** → Layout → Delete
- Existing single-step Up/Down behavior is unchanged

### Notes
- `material-icons-extended` was already a dependency; `SkipPrevious`/`SkipNext` icons available without additional imports
- Follows the same pattern as the existing IPTV `moveGroupToTop`/`moveGroupToBottom` in `IptvRepository.kt`

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CatalogRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CatalogRepository.kt
@@ -824,6 +824,36 @@ class CatalogRepository @Inject constructor(
         return true
     }
 
+    suspend fun moveCatalogToTop(catalogId: String): Boolean {
+        val current = getCatalogs().toMutableList()
+        val visible = current.filter { isVisibleCatalogInSettings(it) }
+        val visibleIndex = visible.indexOfFirst { it.id == catalogId }
+        if (visibleIndex <= 0) return false
+        val currentIndex = current.indexOfFirst { it.id == catalogId }
+        if (currentIndex < 0) return false
+        val moved = current.removeAt(currentIndex)
+        val firstVisibleId = visible.first().id
+        val firstIndex = current.indexOfFirst { it.id == firstVisibleId }
+        current.add(firstIndex.coerceAtLeast(0), moved)
+        saveCatalogs(current)
+        return true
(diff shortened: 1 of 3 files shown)


EierKopZA:
https://github.com/user-attachments/assets/565a2b6a-f82c-429a-8796-ecef541ec81f

PR #233, opened 2026-05-19
Trailers fixes

silentbil:
- Trailer opacity dim — when trailer plays, hero info and card rows fade to transparent so the trailer fills the screen; restores on key press
  - Fixed cold-start flash — overlay no longer flickers back to full alpha when startup sequence swaps the hero item mid-trailer
  - Fixed missing trailers for non-English languages — if Hebrew (or any language) returns no YouTube videos from TMDB, now falls back to English results automatically
  - First key press stops trailer — pressing any D-pad/Enter key while trailer is playing stops the trailer and stays on the current card instead of navigating away; second key press navigates normally
  - Default trailer delay set to 2 seconds, can changed via settings

https://github.com/user-attachments/assets/1c4ce0c5-6432-425a-b3ca-1c8a1c305995


ProdigyV21:
Thanks, this looks good overall. I tested it against latest main and it merges cleanly.

Build checks passed:
- `git diff --check`

One small thing before merge: Trailer Delay can cycle to `0`, and the TV settings row shows that as `Off`, but the home screen treats `0` as “play immediately”, not disabled. Mobile also shows it as `0s`.

So either:
- rename/display `0` as `Instant` / `0s`, or
- if `Off` is intended, make the home trailer not start when delay is `0`.

After that, I think this is good to merge.

silentbil:
> Thanks, this looks good overall. I tested it against latest main and it merges cleanly.
> 
> Build checks passed:
> 
> * `:app:compilePlayDebugKotlin`
> * `:app:compileSideloadDebugKotlin`
> * `git diff --check`
> 
> One small thing before merge: Trailer Delay can cycle to `0`, and the TV settings row shows that as `Off`, but the home screen treats `0` as “play immediately”, not disabled. Mobile also shows it as `0s`.
> 
> So either:
> 
> * rename/display `0` as `Instant` / `0s`, or
> * if `Off` is intended, make the home trailer not start when delay is `0`.
> 
> After that, I think this is good to merge.

Good catch. fixed
0 is one of the options, no more off

PR #234, opened 2026-05-20
fix(network): prevent crash from malformed gzip responses

chillpill244:
## Problem

The app was crashing with:

```
java.io.IOException: gzip finished without exhausting source
  at okio.GzipSource.read(GzipSource.kt:89)
  ...
  at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:153)
  at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
```

This happens when TMDB (via the Supabase proxy) returns a response where extra bytes remain in the network stream after the gzip trailer — likely a Content-Length mismatch or proxy artifact. OkHttp's `GzipSource` (okio) is strict and throws on any trailing bytes.

The crash was fatal because the exception becomes an **undelivered exception**: OkHttp's callback fires on a thread-pool thread just as the Kotlin coroutine is being cancelled (user navigated away), so `continuation.resumeWithException(IOException)` has nowhere to deliver it and falls through to Android's `UncaughtExceptionHandler` → `Process.killProcess()`.

## Fix

Add `lenientGzipInterceptor` as a **network interceptor** in `OkHttpProvider.buildAppClient()`.

It intercepts gzip responses before OkHttp's `BridgeInterceptor` can wrap them with `GzipSource`. It decompresses using Java's `GZIPInputStream`, which is lenient on trailing bytes: it tries to parse them as a concatenated gzip stream, fails silently, and returns. The `Content-Encoding` header is stripped so `BridgeInterceptor` does not attempt a second decompression pass.

## Why a network interceptor and not an application interceptor?

Response processing order in OkHttp:

```
network call → network interceptors → CacheInterceptor → BridgeInterceptor (GzipSource) → application interceptors → Retrofit
```

A network interceptor runs before `BridgeInterceptor` and sees the raw gzip body. An application interceptor would run after `BridgeInterceptor` has already tried (and crashed with) `GzipSource`.

## Testing

Reproduced the `gzip finished without exhausting source` crash via logcat, applied the fix, and confirmed the same code path now completes without error.

## Checklist
- [x] Only `OkHttpProvider.kt` changed — no API, model, or UI impact
- [x] Builds clean (`assembleSideloadDebug`)
- [x] Fallback: if `GZIPInputStream` itself throws, the original response is returned unchanged

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/network/OkHttpProvider.kt
+++ b/app/src/main/kotlin/com/arflix/tv/network/OkHttpProvider.kt
@@ -21,10 +21,13 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
 import java.io.File
+import java.io.IOException
 import java.net.Inet4Address
 import java.net.InetAddress
 import java.net.UnknownHostException
 import java.util.concurrent.TimeUnit
+import java.util.zip.GZIPInputStream
+import okhttp3.ResponseBody.Companion.toResponseBody
 
 /**
  * Provides a configured OkHttpClient instance.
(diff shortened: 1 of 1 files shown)


ProdigyV21:
Thanks, the idea is good and it does compile against latest main.

Checks passed:
- `git diff --check`

One thing I would fix before merge:

In `lenientGzipInterceptor`, if `GZIPInputStream(body.byteStream()).readBytes()` throws, the code returns the original `response`. But the body stream may already be partially consumed by then, so returning the original response is not safe. Retrofit/OkHttp can then receive a half-read body and still fail or parse corrupted data.

Please change the fallback so it does not return a partially consumed response. For example, buffer the raw bytes first and only build a new response from those bytes, or return a controlled error response if lenient decompression fails.

Also consider limiting this to API/proxy JSON responses, because it currently reads every gzip API response fully into memory.

After that, I think this is a useful crash fix.

ProdigyV21:
Merged a hardened replacement in #241.

I did not merge this branch directly because the fallback can return the original response after `GZIPInputStream` may have already consumed part of the body, and the TMDB incident included double-gzipped JSON with only one `Content-Encoding: gzip` header.

#241 keeps the crash fix but buffers compressed bytes before decoding, limits it to gzip JSON responses, caps reads at 16 MiB, decodes repeated gzip layers, and returns a controlled `502` JSON response if decoding fails.

PR #256, opened 2026-05-22
fix(tv): unify settings UI design and fix duplicate titles

Himanth-reddy:
Unifies TV settings screens under a consistent, high-fidelity transparent styling language and fixes duplicate section titles.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/settings/SettingsScreen.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/settings/SettingsScreen.kt
@@ -3596,205 +3596,87 @@ private fun MobileSettingsSubPage(
                 }
             }
             "Plugins & Extensions" -> {
-                MobileSettingsCategory(title = "ADDONS") {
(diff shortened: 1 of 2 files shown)


ProdigyV21:
Can you verify one thing before I merge?

The styling looks good, but from the diff it looks like the mobile settings pages may have lost some management actions:

- Addons: can mobile still remove/delete an addon, or only toggle/add?
- Catalogs: can mobile still move up/down and delete catalogs, or only add/rename?
- TV playlists: can mobile still toggle, reorder, and delete individual playlists, or only edit them / delete all IPTV data?

Maybe I’m reading it wrong, but I want to make sure these controls are not accidentally removed on mobile. If they are still available somewhere, please point me to where/how.

Himanth-reddy:
Good catch! I've updated the `MobileSettingsRow` component to support `onMoveUp`, `onMoveDown`, and `onDelete` actions. I then applied these actions to the mobile settings pages for Addons, Catalogs, and TV Playlists. You can now delete addons/catalogs/playlists and reorder catalogs/playlists just like on the TV UI.

PR #257, opened 2026-05-22
chore(refactor): optimize gson typetoken parsing and extract UI regexes

Himanth-reddy:
### Summary of Changes
* `CloudSyncRepository.kt`: Replaced inline anonymous `TypeToken` class instantiations with static `TypeToken.getParameterized(...)` calls to avoid unnecessary GC churn during synchronization parsing.
* `TraktRepository.kt`: Optimized `Gson` parsing by pre-parameterizing the list type for watchlist items rather than generating anonymous type objects on the fly.
* `StreamRepository.kt`: Upgraded `TypeToken` maps and lists serialization types to use static parameterization for memory safety.
* `ProfileRepository.kt`: Swapped the profile list `TypeToken` initializer to use the static class reference methodology.
* `MediaRepository.kt`: Replaced complex nested `TypeToken` anonymous objects for `List<Map<String, Any?>>` with the equivalent static nested parameterization.
* `SettingsViewModel.kt`: Refactored quality config JSON parsing to use static `TypeToken.getParameterized`.
* `SearchViewModel.kt`: Extracted search pattern `Regex` objects (`LIKE_MATCH_REGEX`, `LIMIT_MATCH_REGEX`) into a companion object `SearchRegexes` to stop repetitive compilation during search queries.

### Why this improves the codebase
* **Performance & Memory Optimization**: By removing anonymous `TypeToken` implementations within loops or hot JSON-parsing paths, the garbage collector (GC) churn and classloader metaspace overhead are significantly reduced. Reusing statically compiled `Regex` objects eliminates redundant compilation times, specifically leading to snappier UI responses during frequent operations like typing in search fields.

### Verification
- [x] Code compiles successfully inside the sandbox environment.
- [x] No new features were added; existing features were fortified and optimized across the modified files.
- [x] Automated tests pass.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
@@ -699,7 +699,7 @@ class CloudSyncRepository @Inject constructor(
         // ── Profiles ──
         val avatarImagesById = root.optJSONObject("profileAvatarImagesById")
         root.optJSONArray("profiles")?.toString()?.takeIf { it.isNotBlank() }?.let { json ->
-            val type = object : TypeToken<List<Profile>>() {}.type
+            val type = TypeToken.getParameterized(List::class.java, Profile::class.java).type
             val profiles: List<Profile> = gson.fromJson(json, type) ?: emptyList()
             val activeProfileId = root.optString("activeProfileId").ifBlank { null }
             if (profiles.isNotEmpty()) {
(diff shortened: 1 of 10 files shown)


ProdigyV21:
This looks fine as an optimization, but it needs a rebase first.

Current main already has the SearchRegexes change from PR253, so this PR conflicts in SearchViewModel.kt and adds a second/duplicate SearchRegexes object. Please rebase onto latest main and keep only one SearchRegexes object/name.

After resolving that, the TypeToken changes look okay and should be safe.

Himanth-reddy:
Rebased onto main and removed the duplicate SearchRegexes object.

PR #270, opened 2026-05-28
Security: add pairing token validation to AI key onboarding

jonahmichael:
## Summary

This PR adds token-based validation to the local AI-key onboarding flow used by the QR setup process.

Previously, the local `/api/key` endpoint accepted unauthenticated submissions from any device able to reach the NanoHTTPD server on the local network.

This change introduces a one-time pairing token that is embedded into the QR onboarding URL and required for successful key submission.

### AiKeyConfigServer

* Added pairing token generation and tracking
* Exposed the active token via `currentPairingToken`
* Added token validation for `/api/key`
* Returns HTTP `403 Forbidden` for missing/invalid tokens
* Added inline comments documenting rationale and future hardening recommendations

### AiKeyWebPage

* Updated Groq/Gemini onboarding pages to:

  * read token from URL query parameter
  * include token in `/api/key` POST payload
* Added inline comments documenting token usage

### SettingsViewModel

* Added pairing token to generated QR onboarding URL using `?t=<token>`

## Security Impact

Before this change:

* Any device on the same LAN could potentially submit API keys to the onboarding endpoint.

After this change:

* Key submission requires possession of the server-generated pairing token.
* The token is automatically supplied through the QR onboarding flow.
* Unauthorized submissions without the token are rejected.

This improves protection against unintended or malicious LAN-based submissions while preserving the current onboarding UX.

## Notes / Future Improvements

Potential future hardening ideas include:

* token expiry or single-use semantics
* localhost-only binding
* HTTPS/TLS support
* clearer invalid-token UI feedback

## Validation

Manually tested:

* valid QR onboarding flow
* successful token-authenticated submission
* rejection of invalid/missing token requests
* successful onboarding UX preservation

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/server/AiKeyConfigServer.kt
+++ b/app/src/main/kotlin/com/arflix/tv/server/AiKeyConfigServer.kt
@@ -2,12 +2,34 @@ package com.arflix.tv.server
 
 import fi.iki.elonen.NanoHTTPD
 import java.io.ByteArrayInputStream
+import java.security.SecureRandom
 
+/**
+ * Local HTTP server used to collect a user-provided AI API key during onboarding.
+ *
+ * Security hardening: require a one-time pairing token to be included with the
+ * POST to `/api/key`. The token is short-lived and displayed on-device (and
+ * encoded into the QR URL by the caller) which reduces the attack surface when
+ * the server is reachable on the local network. For stronger protection consider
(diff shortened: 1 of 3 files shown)


jonahmichael:
Hey @ProdigyV21 ,

It looks like the labels haven't been added yet, so the contribution is not being reflected on my profile.

Could you please add the labels when you get a chance?

Thank you for your help!

PR #271, opened 2026-05-28
Security: replace wildcard CORS with origin allowlist

jonahmichael:
## Summary

This PR replaces permissive wildcard CORS handling with an allowlist-based origin validation approach.

Previously, responses returned:

```http
Access-Control-Allow-Origin: *
```

This allowed requests from arbitrary browser origins.

The new implementation validates the incoming `Origin` header against an environment-configured allowlist and only echoes approved origins.

## Changes Made

* Replaced wildcard `Access-Control-Allow-Origin: *`
* Added request-aware `corsHeaders(req)` helper
* Added environment-driven allowlist support via:

  * `CORS_ALLOWED_ORIGINS`
* Default allowed origins:

  * `https://auth.arvio.tv`
  * `https://arvio.tv`

## Security Impact

This reduces unintended cross-origin access exposure and aligns the edge-function CORS policy with least-privilege principles.

Disallowed origins no longer receive permissive CORS access headers.

## Validation

Manually tested:

* allowed-origin requests
* disallowed-origin requests
* OPTIONS/preflight requests

Verified that:

* allowed origins receive echoed ACAO headers
* disallowed origins do not receive permissive CORS access

## Deployment Notes

Configure allowed origins via:

```env
CORS_ALLOWED_ORIGINS=https://auth.arvio.tv,https://arvio.tv
```

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/server/AiKeyConfigServer.kt
+++ b/app/src/main/kotlin/com/arflix/tv/server/AiKeyConfigServer.kt
@@ -2,12 +2,34 @@ package com.arflix.tv.server
 
 import fi.iki.elonen.NanoHTTPD
 import java.io.ByteArrayInputStream
+import java.security.SecureRandom
 
+/**
+ * Local HTTP server used to collect a user-provided AI API key during onboarding.
+ *
+ * Security hardening: require a one-time pairing token to be included with the
+ * POST to `/api/key`. The token is short-lived and displayed on-device (and
+ * encoded into the QR URL by the caller) which reduces the attack surface when
+ * the server is reachable on the local network. For stronger protection consider
+ * binding to localhost only or serving over TLS; this change implements a
+ * pragmatic one-time token improvement that preserves the previous UX (QR -> phone).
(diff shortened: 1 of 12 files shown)


ProdigyV21:
This looks like a good security improvement overall. I like replacing `Access-Control-Allow-Origin: *` with an allowlist, and the AI key QR token also makes sense.

One thing I want fixed before merging: in `supabase/functions/app-usage-event/index.ts`, `jsonResponse()` calls `corsHeaders(new Request(''))` instead of using the real incoming request. That looks unsafe/suspicious and may break in Deno or return the wrong CORS headers.

Can you change it so `jsonResponse` receives the actual `req`, for example `jsonResponse(req, body, status)`, and then all responses use `corsHeaders(req)`?

After that I think this should be safe to merge, assuming `CORS_ALLOWED_ORIGINS` is set to include the real ARVIO web domains.

jonahmichael:
Heyy!!

I updated jsonResponse() to accept the actual request object and changed all the call sites to pass the real request through.

PR #272, opened 2026-05-28
Refactor: extract reusable settings row composables

jonahmichael:
## Summary

This PR extracts shared settings row composables from `SettingsScreen.kt` into a reusable components file.

The existing settings screen already contained reusable row abstractions (`SettingsRow` and `SettingsToggleRow`) inside the large screen file. This refactor moves those shared primitives into the `ui/components` package without changing behavior or call-site usage.

## Changes Made

* Added:

  * `ui/components/SettingsRows.kt`
* Moved:

  * `SettingsRow`
  * `SettingsToggleRow`
* Updated:

  * `SettingsScreen.kt` imports and references

## Motivation

The goal of this refactor is to:

* reduce `SettingsScreen.kt` size/monolith complexity
* centralize reusable UI primitives
* improve maintainability
* make shared settings components easier to reuse across screens

The implementation intentionally preserves existing behavior and existing call-site usage.

## Validation

Validated with:

* Kotlin error inspection
* compile verification after extraction

No behavioral or UI changes were introduced.


ProdigyV21:
Good idea, but I don’t think this PR is ready yet.

The settings row refactor itself makes sense, but this branch also includes the CORS / AI key security changes from PR271. That should not be in this PR if this one is only meant to extract SettingsRow / SettingsToggleRow.

Also, the new SettingsRows.kt looks like it imports theme values from the wrong package:

com.arflix.tv.ui.screens.settings.ArflixTypography
com.arflix.tv.ui.screens.settings.Pink
com.arflix.tv.ui.screens.settings.TextPrimary
etc.

Those should come from the theme/skin packages, otherwise this may not compile.

Can you clean/rebase this branch so it only contains the SettingsRows refactor, fix the imports, and leave the CORS/AI-key changes in PR271?

jonahmichael:
Thanks for pointing that out. You’re right.

This PR should only contain the SettingsRow / SettingsToggleRow refactor, and the CORS + AI key security changes should stay in PR271. I’ll clean up the branch and rebase so that this pR only includes the settings row extraction changes.

I’ll push an updated version shortly with 

* only the settings row refactor
* corrected imports
* no overlap with PR271

Thanks again for the review

jonahmichael:
Hey @ProdigyV21,

It looks like the labels haven't been added yet, so the contribution is not being reflected on my profile.

Could you please add the labels when you get a chance?

Thank you for your help!

PR #280, opened 2026-05-29
feat(downloads): offline download feature for movies and episodes

chillpill244:
## Summary

Adds end-to-end offline download support for mobile users. Downloads are restricted to user-controlled local servers (Jellyfin, Emby, Plex) to comply with Play Store policy.

### Core infrastructure
- **Room database** (`DownloadEntity`, `DownloadDao`, `ArflixDatabase`) — tracks download status, progress percentage, local file path, and artwork metadata (poster, backdrop, TMDB ID, season/episode numbers)
- **`DownloadsRepository`** — provides insert, query, pause/resume/cancel/delete/retry operations; deletes the on-disk file when a download is removed
- **`DownloadWorker`** — `CoroutineWorker` that streams the file over OkHttp with `Range` header support for true pause/resume, posts a persistent foreground notification with live progress, and handles Android 12+ foreground service restrictions gracefully

### Downloads UI
- **`DownloadsTab`** — horizontal card rows split into Movies and Series sections; artwork respects the user's existing poster/backdrop `CardLayoutMode` setting
- **`DownloadActionsSheet`** — bottom sheet offering pause, resume, cancel, delete, and retry — actions available depend on the current `DownloadStatus`
- **`DownloadedEpisodesScreen`** — per-series episode list, navigable from the Downloads tab
- Series cards show a completed/total episode badge; long-pressing opens a "Delete all downloads" sheet for the whole series

### Entry points
- **Bottom bar** gains a Downloads tab on mobile when at least one download exists
- **Details screen** shows a Download / Downloading `x%` / Downloaded / Paused / Retry button below the primary Play button, styled with a subtle glass outline to match the Trailer button; tapping a completed download prompts deletion
- **Context menu** on watchlist and catalog cards gains a Download option

### Player
- Chromecast cast button is hidden and the Cast handler is no-op'd for offline (`file://`) streams — Chromecast devices cannot access local device storage

### Watchlist / Downloads tab navigation
- Replaced the Material `TabRow` with an **Apple TV-style animated pill segmented control** — smooth background and text color transitions between tabs
- Tab selection is now preserved with `rememberSaveable` so returning from Details always restores the tab the user was on

### Additional improvements
- **Stream headers forwarding**: Per-stream proxy headers (`Referer`, `Authorization`, etc.) from Stremio addon `behaviorHints.proxyHeaders` are forwarded to `DownloadWorker`, matching the credentials ExoPlayer uses during playback
- **Session-cookie detection**: The download stream picker runs a HEAD probe on selection; streams that respond with `Set-Cookie` are shown as unavailable for download rather than silently failing
- **IPTV file size resolution**: Two-stage probe — HEAD first, then `Range: bytes=0-0` GET fallback for IPTV/Xtream Codes servers that omit `Content-Length` on HEAD; resolved size appears next to the Download button
- **Subtitle preferences**: Download picker honours `default_subtitle` / `secondary_subtitle` from Settings, filtering the list and auto-selecting a matching track
- **Movie size badge**: Completed movie cards in the Downloads tab show a file size overlay (e.g. "2.1 GB") in the top-left corner
- **`formatBytes` extracted** to `com.arflix.tv.util.DownloadUtils`; subtitle matching helpers extracted to `com.arflix.tv.util.SubtitleLanguageMatcher` — removes identical private copies across multiple files

### Play Store compliance
- **Download picker filtered to local servers only** — `DownloadSheet` only surfaces streams where `addonId == "home_server"` (Jellyfin, Emby, Plex). The user sees "Downloads are only available for your local server (Jellyfin, Emby, or Plex)" if no local server streams are present.

## Test plan
- [x] Shows watchlist icon in navigation by default 
- [x] Watchlist screen has two tabs - Watchlist/Downloads
- [x] Download a movie from Jellyfin → verify progress notification → tap play after completion
- [x] Download a series episode from Jellyfin → navigate to Downloaded Episodes screen
- [x] Shows downloads icon in navigation when downloads are present
- [x] Shows download progress in navigation when movie/episodes are downloading
- [x] Pause mid-download → resume → verify it continues from where it left off (not restarted)
- [x] Delete a download → verify file is removed from device storage
- [x] Verify Chromecast button is absent on the player screen for a downloaded file
- [x] Switch to watchlist tab → navigate to Details by clicking on any media → back → confirm tab stays on Watchlist
- [x] Open download sheet for a Stremio/IPTV source → verify "local server only" message appears and no streams are listed
- [x] Set a preferred subtitle language in Settings — verify matching subtitle is pre-selected in the download picker
- [x] Complete a movie download — verify size badge appears on the card in Downloads tab

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -324,6 +324,11 @@ dependencies {
     implementation("androidx.hilt:hilt-work:1.1.0")
     ksp("androidx.hilt:hilt-compiler:1.1.0")
 
+    // Room for download persistence
+    implementation("androidx.room:room-runtime:2.7.1")
+    implementation("androidx.room:room-ktx:2.7.1")
+    ksp("androidx.room:room-compiler:2.7.1")
+
     // Profile installer for baseline profiles
     implementation("androidx.profileinstaller:profileinstaller:1.3.1")
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -7,6 +7,9 @@
     <!-- Permissions -->
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC" />
 
     <!-- TV Features (not required so the same APK installs on phones/tablets) -->
     <uses-feature
@@ -128,6 +131,11 @@
                 <action android:name="${applicationId}.INSTALL_COMPLETE" />
             </intent-filter>
         </receiver>
+
+        <service
+            android:name="androidx.work.impl.foreground.SystemForegroundService"
+            android:foregroundServiceType="dataSync"
+            android:exported="false" />
 
     </application>
 
--- a/app/src/main/kotlin/com/arflix/tv/ArflixApplication.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ArflixApplication.kt
@@ -80,6 +80,16 @@ class ArflixApplication : Application(), Configuration.Provider, ImageLoaderFact
         super.onCreate()
         instance = this
 
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O) {
+            val channel = android.app.NotificationChannel(
+                com.arflix.tv.worker.DownloadWorker.NOTIFICATION_CHANNEL_ID,
+                "Downloads",
+                android.app.NotificationManager.IMPORTANCE_LOW
+            )
+            getSystemService(android.app.NotificationManager::class.java)
+                .createNotificationChannel(channel)
+        }
+
         // OkHttpProvider.init(context) just stashes the app context; it does
         // not build the OkHttpClient. Safe to keep on the main thread — it's
         // a single volatile assignment.
--- a/app/src/main/kotlin/com/arflix/tv/MainActivity.kt
+++ b/app/src/main/kotlin/com/arflix/tv/MainActivity.kt
@@ -35,6 +35,7 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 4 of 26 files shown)


ProdigyV21:
I really like the idea, great pr! and offline downloads would be a strong feature, but I’m worried about Play Store risk.

I want ARVIO to stay the same version on Play Store and GitHub. I don’t want a “Play Store version” and a separate “GitHub version”. Because of that, we have to follow the stricter Play Store rules.

The problem is that this PR allows downloading video files from general streaming sources like IPTV/Stremio/Debrid/unknown links. Even if the user adds the source themselves, Google can still see the app as enabling users to save copyrighted movies/episodes locally without permission. Play Store policy is very strict about apps that stream or download copyrighted content without authorization, and this could risk rejection or even app suspension.

For self-owned sources like Jellyfin/local media it is much safer, because the user controls the server/content. But “download any stream” is too broad for a Play Store app.

So I don’t think we should merge this as-is. If we keep it, it should be limited to clearly self-hosted/owned sources only, or redesigned as temporary playback cache instead of a visible offline download library. Otherwise I think it is better to skip full offline downloads so ARVIO stays safe for Play Store. If you have any other suggestions about this let me know, you could also look into the policies more, it really falls into a green or grey zone we could include this feature.

chillpill244:
Hey @ProdigyV21, addressed this with two changes:

1. The Download button on the details screen is now only shown when the user has a home server (Jellyfin, Emby, or Plex) configured — so users without a local server never see the option at all.
2. As a defense-in-depth measure, the download stream picker also filters to local server streams only (`addonId == "home_server"`), so even if the button were somehow reached, IPTV/Stremio/Debrid streams would never be downloadable.

ProdigyV21:
Offline downloads from user-owned Jellyfin/Emby/Plex servers would be useful and much safer than generic stream downloads.

But PR280 is not ready to merge right now. Apologies, because I overlooked this pr it no longer merges into latest main. I get conflicts in AppNavigation.kt, ContextMenu.kt, and DetailsScreen.kt, so it needs a rebase/manual conflict resolution first.

After rebasing, please verify both builds compile and that downloads are still restricted to home_server/local-server streams only. Once that is done, this could be a strong feature and will be merged.

chillpill244:
> Offline downloads from user-owned Jellyfin/Emby/Plex servers would be useful and much safer than generic stream downloads.
> 
> But PR280 is not ready to merge right now. Apologies, because I overlooked this pr it no longer merges into latest main. I get conflicts in AppNavigation.kt, ContextMenu.kt, and DetailsScreen.kt, so it needs a rebase/manual conflict resolution first.
> 
> After rebasing, please verify both builds compile and that downloads are still restricted to home_server/local-server streams only. Once that is done, this could be a strong feature and will be merged.

@ProdigyV21  Done, updated. Verified download buttons shows up only when there is home server setup and downloading movies/series from Home server works. Other sources don't show up in the download options.

ProdigyV21:
This pr slipped my mind, my bad. Can you rebase it to solve the conflict? Then I will merge.

philstanway-oss:
@chillpill244 where are we with this?

PR #297, opened 2026-06-03
Telegram integration + APK split

silentbil:
Native Telegram Integration + ABI Split APKs
  
  Telegram Integration
  
  Adds a built-in Telegram source that searches the user's joined channels and groups for video files whenever a movie or episode is selected — appearing alongside regular addon sources in the stream selector. No third-party service or external addon required.

  Authentication
  - QR code flow via TDLib (authorizationStateWaitQrCodeConfirmation) — scan with the Telegram phone app, no typing on TV
  - 2FA password support
  - Session persists across app restarts

  Streaming
  - Local Ktor CIO HTTP server proxies TDLib file downloads to ExoPlayer with Range header support for seeking

  Settings Screen
  - Shows connected account name or "Not connected"
  - Per-channel toggle to exclude channels from search
  - Disconnect button wipes session and TDLib database

  Source Card
  - Quality parsed from filename (4K, 1080p, 720p, etc.) including OCR typos (1o8o, 216o)
  - File size displayed

  Native Libraries
  - TDLib prebuilt binaries downloaded automatically on first build via downloadTdlibNatives Gradle task
  - Cached — re-download only on clean checkout

  ---
  ABI Split APKs (Sideload)

  assembleSideloadRelease now produces per-ABI APKs instead of one universal APK, so users download only the native library for their device architecture.

  Output APKs

  ┌──────────────────────────────────────┬────────┬────────────────────────────────────────────────────┐
  │                 File                 │  Size  │                        Use                         │
  ├──────────────────────────────────────┼────────┼────────────────────────────────────────────────────┤
  │ app-sideload-arm64-v8a-release.apk   │ ~49 MB │ Modern 64-bit streamers (Shield, Fire TV 4K, etc.) │
  ├──────────────────────────────────────┼────────┼────────────────────────────────────────────────────┤
  │ app-sideload-armeabi-v7a-release.apk │ ~42 MB │ Older 32-bit streamers                             │
  ├──────────────────────────────────────┼────────┼────────────────────────────────────────────────────┤
  │ app-sideload-universal-release.apk   │ ~70 MB │ All devices — use when ABI is unknown              │
  ├──────────────────────────────────────┼────────┼────────────────────────────────────────────────────┤
  │ app-sideload-x86-release.apk         │ —      │ Emulator only, do not distribute                   │
  ├──────────────────────────────────────┼────────┼────────────────────────────────────────────────────┤
  │ app-sideload-x86_64-release.apk      │ —      │ Emulator only, do not distribute                   │
  └──────────────────────────────────────┴────────┴────────────────────────────────────────────────────┘

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -172,13 +172,13 @@ android {
         }
     }
 
-    // Per-ABI APKs for sideload distribution. Disabled for AAB/Play Store builds because
-    // Google Play handles ABI splits in the cloud via Dynamic Delivery.
-    // x86/x86_64 splits are for emulator/debug only (those .so files live in src/debug/jniLibs).
-    val buildingAppBundle = gradle.startParameter.taskNames.any { it.contains("bundle", ignoreCase = true) }
+    // Per-ABI APKs for sideload distribution only.
+    // Play builds are distributed via AAB (Google Play handles splitting) so splits are not needed.
+    // x86/x86_64 splits are emulator/debug use only.
+    val isSideloadBuild = gradle.startParameter.taskNames.any { it.contains("sideload", ignoreCase = true) }
     splits {
         abi {
-            isEnable = !buildingAppBundle
+            isEnable = isSideloadBuild
             reset()
             include("armeabi-v7a", "arm64-v8a", "x86", "x86_64")
             isUniversalApk = true
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -401,7 +401,7 @@ tasks.register("downloadTdlibNatives") {
     onlyIf { !marker.exists() }
 
     doLast {
-        val base = "https://github.com/FaiBah/tdlib-android-prebuilt/releases/download/v1.8.64-e0943d068-Java"
+        val base = "https://github.com/FaiBah/tdlib-android-prebuilt/releases/download/v1.8.64-e0943d0-Java"
 
         fun fetch(urlStr: String, dest: File) {
             dest.parentFile.mkdirs()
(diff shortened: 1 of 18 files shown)


ProdigyV21:
I tested the PR against current main: it merges cleanly, `git diff --check` passes, and both `:app:compilePlayDebugKotlin` and `:app:compileSideloadDebugKotlin` compile.

Before merge, I think these 3 things need to be fixed/clarified:

1. Telegram source should be sideload-only or disabled in Play builds.
Right now it is in `src/main`, so it ships with the Play Store build too. Since it searches Telegram chats/channels for video files and streams them, this may be a Play Store policy risk. I don’t want to risk the Play version getting flagged.

2. TDLib binaries need provenance/checksum clarity.
The PR commits TDLib generated Java and native `.so` files, and also has a Gradle task downloading prebuilt binaries from a third-party GitHub release. Please add clear notes about where these binaries come from, their license/provenance, and ideally checksum verification for the download.

3. ABI split should be limited to sideload if that is the goal.
The PR says ABI split APKs are for sideload, but the Gradle config currently enables ABI splits for any APK build unless it is an app bundle. Please gate this to sideload variants only, so Play/debug APK behavior is not changed unexpectedly.

silentbil:
@ProdigyV21 thanks for reviewing

1.
>this may be a Play Store policy risk. 

 Actual risk is low because:
  - TDLib is used by many legitimate apps on Play Store (it's the official Telegram client library)
  - You're not scraping, bypassing DRM, or accessing anything the user doesn't own — the user authenticates with their own account and searches their own channels
  - Google's automated scanning looks for known malware patterns, not "this app talks to Telegram"

2. fixed
3. fixed

PR #306, opened 2026-06-03
Implement bounded LruCache in MediaRepository to prevent OOM crashes

### LINKED ISSUE #299: Unbounded In-Memory Cache Growth
Bug Report | Memory Leak Critical

Description
Multiple caches in MediaRepository (detailsCache, castCache, similarCache, logoCache, etc.) use mutableMapOf without size limits or TTL-based eviction. Long user sessions with browsing can exhaust device heap memory on TV devices with limited RAM (384-512 MB).

Steps to Reproduce
Install app on TV device with constrained memory
Browse catalog continuously for 30+ minutes
View details for 100+ different media items
Memory usage grows unbounded; app crashes or becomes sluggish
Expected Behavior
Caches should have bounded size and TTL-based eviction

Actual Behavior
Memory usage grows linearly with items accessed; no eviction policy

Environment
Platform: Android TV (low-memory devices)
Component: Media Repository
app/src/main/kotlin/com/arflix/tv/data/repository/MediaRepository.kt (line 110-123)
Possible Fix
Use LruCache or implement a TTL-based eviction with max size limits.

annrose2277-glitch:
🐛 Background / Root Cause
 The MediaRepository previously utilized eight standard mutableMapOf instances for caching media data (details, cast, reviews, etc.). Because these maps lacked size limits or Time-To-Live (TTL) eviction policies, they accumulated data indefinitely. On low-memory TV environments (typically 384-512 MB RAM), extended browsing sessions resulted in memory exhaustion and application crashes.
closes #299 
🛠️ Solution
Replaced all unbounded mutableMapOf declarations with android.util.LruCache to enforce strict memory boundaries. When a cache reaches its assigned capacity, it now automatically evicts the least recently used entries.

Memory limits applied based on data weight:

detailsCache: Max 200 items

castCache: Max 150 items

similarCache: Max 150 items

logoCache: Max 300 items

reviewsCache: Max 100 items

watchProvidersCache: Max 150 items

seasonEpisodesCache: Max 200 items

imdbRatingCache: Max 500 items

📁 Files Modified
app/src/main/kotlin/com/arflix/tv/data/repository/MediaRepository.kt

✅ Impact & Benefits
Stability: Directly prevents Out-Of-Memory (OOM) crashes during prolonged viewing sessions (1-2+ hours).

Predictability: Ensures a stable, capped memory footprint for the app, leaving sufficient overhead for the TV OS and heavy video playback operations.

Zero UI Impact: Caching logic remains the same; only the underlying storage mechanism was swapped.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/MediaRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/MediaRepository.kt
@@ -116,15 +116,20 @@ class MediaRepository @Inject constructor(
     @Volatile private var homeCategoriesFetchedAt = 0L
     private val HOME_CATEGORIES_CACHE_MS = 120_000L // 2 minutes
 
-    private val detailsCache = LruCache<String, CacheEntry<MediaItem>>(200)
+    private fun <K, V> createBoundedCache(maxSize: Int): MutableMap<K, V> {
+        return object : java.util.LinkedHashMap<K, V>(maxSize, 0.75f, true) {
+            override fun removeEldestEntry(eldest: MutableMap.MutableEntry<K, V>?): Boolean = size > maxSize
+        }
+    }
+    private val detailsCache = createBoundedCache<String, CacheEntry<MediaItem>>(200)
     private val fullDetailsCacheKeys = mutableSetOf<String>()
-    private val castCache = LruCache<String, CacheEntry<List<Cast>>>(150)
-    private val similarCache = LruCache<String, CacheEntry<List<MediaItem>>>(150)
-    private val logoCache = LruCache<String, CacheEntry<String>>(300)
-    private val reviewsCache = LruCache<String, CacheEntry<List<Review>>>(100)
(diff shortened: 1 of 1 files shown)


ProdigyV21:
Good idea, but this implementation cannot be merged yet because it does not compile.

I tested both builds and they fail:

./gradlew :app:compilePlayDebugKotlin
./gradlew :app:compileSideloadDebugKotlin
Main problems:

Cast and WatchProvider do not exist in the current codebase.
Existing cache helpers expect a normal Kotlin Map, but android.util.LruCache is not a Map, so calls like getFromCache(...) and containsKey break.
Some cache types were changed incorrectly, for example nullable cache values became non-null.
The idea is good: bounded caches would help prevent memory growth on low-memory TVs. But it needs to be implemented in a way that matches the current MediaRepository cache helpers. A safer approach may be either adding proper LruCache helper functions or using a bounded LinkedHashMap wrapper that still behaves like the current maps.

annrose2277-glitch (commented):
changes made

PR #307, opened 2026-06-03
Memory Leak in PlaybackCookieJar

### LINKED ISSUE #301: Cookie Jar Memory Leak
Bug Report | Memory Leak

Description
PlaybackCookieJar stores cookies indefinitely in cookiesByHost map. Expired cookies are not properly cleaned up, causing memory accumulation over time with many streaming sources.

Steps to Reproduce
Use multiple streaming sources with HTTP cookies
Run app for several hours
Monitor memory usage
Memory grows due to stale cookies
Expected Behavior
Expired cookies should be cleaned up automatically

Actual Behavior
All cookies remain in memory indefinitely

Environment
Platform: Android TV
Component: HTTP Cookie Management
app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt (line 4979-5009)
Possible Fix
Add automatic cleanup of expired cookies in a background task.

annrose2277-glitch:
During extended playback sessions (3+ hours) with multiple streaming sources, the HTTP layer's PlaybackCookieJar accumulated cookies indefinitely. The underlying cookiesByHost map lacked a mechanism to clean up expired cookies. Over time, this unbounded memory growth caused severe memory pressure, resulting in app slowdowns and eventual crashes on low-memory TV devices.
 closes #301 

🛠️ Solution
Refactored the PlaybackCookieJar class to enforce strict cookie expiration and map cleanup:

Active Eviction on Save (saveFromResponse): The logic now explicitly filters out expired cookies before saving new ones and merges the valid sets. If a host no longer has any valid cookies after this process, its key is completely removed from the map.

Active Eviction on Load (loadForRequest): The jar now purges any expired cookies from the map before returning the valid list to the HTTP client. Empty host entries are dynamically removed during the load process as well.

📁 Files Modified
app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt

✅ Impact & Benefits
Stable Memory Footprint: Eradicates the infinite memory bloat during prolonged viewing sessions.

Improved Performance: Reduces garbage collection (GC) churn, preventing UI stutters and sluggishness on low-RAM TV hardware.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt
@@ -4996,14 +4996,17 @@ private class PlaybackCookieJar : CookieJar {
         val current = cookiesByHost[host]?.toMutableList() ?: mutableListOf()
         val now = System.currentTimeMillis()
 
+        current.removeAll { it.expiresAt <= now }
+
         cookies.forEach { cookie ->
-            if (cookie.expiresAt <= now) return@forEach
-            current.removeAll { existing ->
-                existing.name == cookie.name &&
-                    existing.domain == cookie.domain &&
-                    existing.path == cookie.path
+            if (cookie.expiresAt > now) {
+                current.removeAll { existing ->
+                    existing.name == cookie.name &&
+                        existing.domain == cookie.domain &&
+                        existing.path == cookie.path
+                }
+                current.add(cookie)
             }
-            current.add(cookie)
         }
 
         if (current.isEmpty()) {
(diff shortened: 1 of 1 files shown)


ProdigyV21:
Looks good and useful. The cookie cleanup makes sense and should help avoid memory buildup during long playback sessions.

I checked it and both builds pass:

./gradlew :app:compilePlayDebugKotlin
./gradlew :app:compileSideloadDebugKotlin
Only small thing: git diff --check fails because of trailing whitespace on two blank lines in PlayerScreen.kt. Can you clean those two whitespace lines, then this should be safe to merge.

annrose2277-glitch (commented):
changes made

PR #329, opened 2026-06-06
Fix/hebrew localization

test01203:
🐛 Bug Fixes & Improvements
Global RTL Support: Forced RTL layout direction globally across the app when Hebrew (iw or he) is selected as the app language.
Player UI Excluded from RTL: Explicitly wrapped PlayerScreen in a CompositionLocalProvider with LayoutDirection.Ltr. This ensures that the video player controls, scrubbers, and icons remain in their standard Left-to-Right orientation, preventing confusing reversed playback controls.
TMDB API Localization: Added a custom OkHttp Interceptor (TmdbInterceptor) to inject language=he into all TMDB API requests when the app is set to Hebrew.
Locale Code Mapping: Added a mapping in MediaRepository to convert Android's legacy iw locale code to the standard he required by TMDB for proper metadata and poster fetching.
Category Translations: Added localized strings for the dynamically generated "Top 10 Movies/Shows Today" rows when Hebrew is active.
Text Readability (Drop Shadows): Added Shadow parameters to text components in HomeScreen, MediaCard, and ContinueWatchingCard. This ensures that category titles and metadata text below posters remain clearly visible and readable even when placed over bright or white background art.
📸 UI Changes
D-Pad navigation now correctly moves left/right visually according to the RTL layout on the Home Screen.
Media titles and years now have a subtle black drop-shadow for improved contrast.


ProdigyV21:
I checked PR329. It merges cleanly and both Play + Sideload Kotlin compiles pass.

The feature is useful, especially for Hebrew users, but I would not merge yet because there is one real behavior issue:

In DetailsViewModel, plugin scrapers can keep isLoadingStreams=true forever if normal addons finish with no streams and plugin scrapers also return no results. The plugin job clears pluginScrapersLoading in finally, but it does not clear isLoadingStreams. So the stream selector can stay stuck in a loading state even when all searching is done.

Please set isLoadingStreams=false when the plugin scraper job finishes and there are no remaining streams/loading jobs.

Also please clean the trailing whitespace from git diff --check before merge.

After those two fixes, I think this is safe and useful to merge.

test01203:
Thanks for the review! I've pushed a new commit addressing all your points:

1. Fixed isLoadingStreams logic: Updated the finally block of the pluginScraperJob in DetailsViewModel.kt. It now correctly evaluates whether all other stream-fetching jobs (loadStreamsJob, vodAppendJob, homeServerAppendJob) have finished and clears isLoadingStreams accordingly, preventing the infinite loading state.
2. Cleaned Trailing Whitespaces: Removed all the trailing whitespaces that showed up in git diff --check.
3. Hero UI Improvement (Bonus): I also wrapped the HeroSection in HomeScreen.kt with a LayoutDirection.Ltr provider. Previously, in RTL mode, the hero text (title and synopsis) was mirrored to the right side of the screen. This caused it to obscure the main character/faces in most background images (since backdrops usually place characters on the right). Now it stays neatly on the left side, which significantly improves readability for Hebrew users!
Everything compiles cleanly. Let me know if it's good to merge! 🚀

ProdigyV21:
Sorry this pr slipped my mind. Can you resolve the conflicts? Then I can merge.

test01203:
> Sorry this pr slipped my mind. Can you resolve the conflicts? Then I can merge.

fix

PR #359, opened 2026-06-11
fix: Restore FEATURE_PLUGINS_ENABLED and sideload plugin dependencies

test01203:
Hey , it looks like in one of the recent commits (598df332) the plugin system dependencies and the FEATURE_PLUGINS_ENABLED flags were accidentally removed from app/build.gradle.kts (probably during a merge conflict resolution). This caused the Plugins menu to disappear entirely from the settings. This PR restores them so the feature works again

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -153,6 +153,7 @@ android {
 
     packaging {
         resources {
+            excludes += "/META-INF/versions/9/OSGI-INF/MANIFEST.MF"
             excludes += setOf(
                 "/META-INF/{AL2.0,LGPL2.1}",
                 "/META-INF/LICENSE*",
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -20,7 +20,7 @@ plugins {
 
 android {
     namespace = "com.arflix.tv"
-    compileSdk = 35
+    compileSdk = 36
 
     flavorDimensions += "distribution"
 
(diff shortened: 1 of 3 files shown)


silentbil:
Hey, tested this branch locally — build passes and installs fine, but the app crashes on every launch with:
  ▎
  ▎ java.lang.NoSuchMethodError: No virtual method addLast(LockFreeLinkedListNode;)V
  ▎ in class kotlinx.coroutines.internal.LockFreeLinkedListHead (classes22.dex)
  ▎
  ▎ One of the new sideload deps (likely CloudStream3 or NiceHttp) bundles an older kotlinx-coroutines-core internally. That older class wins in the dex merge and is missing addLast() which the app expects at runtime — instant crash before anything renders.
  ▎
  ▎ Fix: force coroutines to 1.7.3 in the resolution strategy so it overrides the transitive version:
  ▎
  ▎ configurations.all {
  ▎     resolutionStrategy {
  ▎         force("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.3")
  ▎         force("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.7.3")
  ▎     }
  ▎ }
  ▎
  ▎ Or exclude coroutines from the conflicting dep:
  ▎ add("sideloadImplementation", "com.github.recloudstream.cloudstream:library-android:v4.7.0") {
  ▎     exclude(group = "org.jetbrains.kotlinx", module = "kotlinx-coroutines-core")
  ▎     exclude(group = "org.mozilla", module = "rhino")
  ▎ }

silentbil:
fixed on another PR
https://github.com/ProdigyV21/ARVIO/pull/362

Issue #251, opened 2026-05-21
[Enhancement] Add Gemini Flash-Lite 3.1


Akila3:
1. Please add the Gemini Flash-Lite v3.1 model as a selectable option in the AI Translation function.
Alternatively, adding a single entry called “Flash-Lite-Latest”.
That always points to the newest Gemini Flash-Lite model would be even more convenient and future-proof.

2. Add advanced settings, to allow power users to manually configure key generation parameters: Temperature, Top-P (nucleus sampling) and the reasoning/thinking token budget (e.g., 512, 1024, 2048, etc.). This would give power users greater control and flexibility, which is very needed when there is no surrounding context being submitted to the AI.

silentbil:
Thanks @Akila3 i will look into

Akila3:
Any update?

mvanhorn:
Implemented in https://github.com/ProdigyV21/ARVIO/pull/363.

silentbil:
@mvanhorn see my comments in your PR.
u added 2.5 flash lite instead of 3.1

Akila3:
@mvanhorn are we adding Flash 3.x family models?

Himanth-reddy:
This feature has been implemented in ARVIO.

ARVIO's AI Subtitle Translation engine uses `gemini-3.5-flash-lite` for high-throughput, low-cost subtitle translation. The existing profile DataStore and Cloud Sync configuration routes directly to `gemini-3.5-flash-lite` while maintaining backward compatibility for saved settings.

Closing this issue as completed.

PR #363, opened 2026-06-12
feat: add Gemini Flash-Lite as a selectable AI subtitle-translation model

mvanhorn:
## Summary

Gemini Flash-Lite is now a selectable model for AI subtitle translation. The Gemini request path is parameterized by model id instead of baking `gemini-2.5-flash` into the base-URL constant, and the new `GEMINI_FLASH_LITE` enum value routes through the same `translateGemini` implementation with `gemini-2.5-flash-lite`.

## Why this matters

#251 asked for Flash-Lite as an option for faster, cheaper subtitle translation; @silentbil acknowledged it on 2026-05-28 and the reporter pinged again on 2026-06-11 with no PR in sight. This covers the model-addition half of the issue; the advanced generation settings (temperature, top-p, thinking budget) are a separate piece and stay open. Persistence needs no changes since the enum is stored by name with a safe fallback, so existing users see no migration.

## Testing

Unit test covers the enum round-trip and the model-id mapping; the settings screen's model label, disclaimer text, and model dialog all render the new option through the extended exhaustive `when` blocks.

Refs #251

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/player/SubtitleAiModel.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/player/SubtitleAiModel.kt
@@ -9,6 +9,6 @@ enum class SubtitleAiModel {
         get() = when (this) {
             GROQ_LLAMA_70B -> null
             GEMINI_FLASH_25 -> "gemini-2.5-flash"
-            GEMINI_FLASH_LITE -> "gemini-flash-lite-latest"
+            GEMINI_FLASH_LITE -> "gemini-2.5-flash-lite"
         }
 }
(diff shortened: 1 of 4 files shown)


silentbil:
Hi @mvanhorn 
This [request](https://github.com/ProdigyV21/ARVIO/issues/251) was for gemini 3.1 flash lite
while this PR just added a 2.5 flash lite.
I compared 2.5 vs 2.5 lite in the past. and 2.5 gave better quality at almost same price.
i wanted to check out 3.1 flash lite which could make things better, maybe change your pr to use it

silentbil:
inactivity

Akila3:
@mvanhorn are we adding Flash 3.x family models?

mvanhorn:
The PR predates the 3.x family - it added 2.5 Flash Lite since that's what was selectable at the time. Happy to resubmit it targeting gemini-3.1-flash-lite (and the 3.x family generally) if the maintainers want it; the plumbing is the same, only the model ids change.

PR #421, opened 2026-07-01
feat: implement Discord crash reporting flow for TV and mobile

Himanth-reddy:
### Description
Implements an in-app crash reporting recovery screen and web bridge for seamless submission of bug reports to Discord.

### Features
- **Uncaught Exception Handler & Sentry Interception**: Captures crash diagnostics (, error message, release version, timestamp) and persists them before application exit.
- **Isolated Crash UI Process ()**: Displays a dedicated  running in its own process.
- **Android TV Support**: Renders a clean QR code pointing to  where users can scan with their phone camera to copy details and jump into Discord.
- **Mobile Support**: 1-click button to copy markdown crash diagnostics (including direct Sentry issue search query URL) to clipboard and launch Discord.
- **Web Bridge ()**: A Next.js landing page that formats crash search parameters and provides a 1-click copy & Discord redirect flow.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -69,6 +69,14 @@
 
         </activity>
 
+        <!-- Crash Report Activity -->
+        <activity
+            android:name=".ui.screens.crash.CrashReportActivity"
+            android:configChanges="orientation|screenSize|smallestScreenSize|screenLayout|keyboard|keyboardHidden"
+            android:exported="false"
+            android:process=":crash"
+            android:theme="@style/Theme.ArflixTV" />
+
         <!-- Disable default WorkManager initialization for Hilt -->
         <provider
             android:name="androidx.startup.InitializationProvider"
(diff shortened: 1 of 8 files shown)


ProdigyV21:
1. The Sentry crash ID can be stale/wrong. The crash handler reads `last_crash_id` before Sentry has created the new crash event, so the Discord report may link to an old crash or a random ID.

2. The QR uses `https://arvio.tv/report`, but the new page is added under the web app. Please confirm that exact URL is actually deployed/redirected, otherwise TV users will scan the QR and land on a missing page.

3. Starting `CrashReportActivity` from the uncaught exception handler and then immediately calling the default crash handler is risky. The app process may die before the crash UI appears. Safer would be: save crash info, let the app close, then show the report screen on next launch.

4. The TV crash screen needs proper D-pad focus so users can use Restart ARVIO with only the remote.

Also small cleanup: `web/app/report/page.tsx` has trailing whitespace.

So: feature is good, but please fix these before merge.

PR #424, opened 2026-07-02
fix: resolve profile deletion sync bug and homescreen snap-back

### LINKED ISSUE #423: Profile
I'm unable to delete profile

Himanth-reddy:
### Summary
Fixes a critical profile sync bug where deleting a profile would cause the UI to snap back to the homescreen, and the deleted profile would reappear when re-opening switch profiles.

### Root Cause Analysis
During profile deletion, three mechanisms improperly blocked or reverted the local deletion:
1. **Server-Side Snapshot Rejection**: Both the Netlify cloud function (`account-sync-push.js`) and Supabase SQL function (`save_account_sync_payload`) evaluated incoming cloud snapshots against existing cloud storage using a heuristic that checked if `existing.profileCount > incoming.profileCount`. Because deleting a profile reduces the total profile count, legitimate deletions were rejected with `existing_snapshot_is_richer`.
2. **Client-Side Remote Restore**: When `pushToCloud(force = true)` ran after deletion, `CloudSyncRepository.shouldRestoreRemoteBeforePush()` evaluated the local snapshot against the un-updated remote payload. Seeing that the remote snapshot had more profiles, it blocked the push and applied the remote snapshot over local storage.
3. **Homescreen Snap-Back**: When `applyCloudPayload` restored the old cloud snapshot into `profilesDataStore`, `ProfileSelectionScreen` detected an unexpected change to `activeProfile` while `navigateTriggered` was active, instantly calling `onProfileSelected()` and navigating back to the homescreen.

### Changes Made
- **Android Client (`CloudSyncRepository.kt`)**:
  - Updated `shouldRestoreRemoteBeforePush` so configured local snapshots (`restoreRank >= 40`) are never overridden by remote cloud snapshots before a push.
  - Updated `pushToCloud(force = true)` to explicitly pass `allowRemoteRestoreBeforePush = false`.
- **Netlify Cloud Backend (`_backend.js`)**:
  - Updated `isExistingSnapshotRicher` so configured incoming payloads (`incomingRank >= 40`) are accepted even when profile count or coverage decreases due to legitimate deletion.
- **Supabase SQL Migration (`20260618_account_delta_sync.sql`)**:
  - Updated `save_account_sync_payload` to align with the improved heuristic (`v_incoming_rank < 40 and v_existing_rank >= 40`).

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
@@ -140,19 +140,9 @@ class CloudSyncRepository @Inject constructor(
     private fun shouldRestoreRemoteBeforePush(localPayload: String, remotePayload: String): Boolean {
         val localRank = accountSyncPayloadRestoreRank(localPayload)
         val remoteRank = accountSyncPayloadRestoreRank(remotePayload)
-        val localProfileCount = cloudPayloadProfileCount(localPayload)
-        val remoteProfileCount = cloudPayloadProfileCount(remotePayload)
 
-        if (
-            localProfileCount != null &&
-            remoteProfileCount != null &&
-            remoteProfileCount > localProfileCount &&
-            remoteRank >= localRank
-        ) {
-            return true
-        }
-
-        return remoteRank >= 70 && localRank < 70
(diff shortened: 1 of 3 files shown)


Himanth-reddy:
Solves #423

PR #433, opened 2026-07-02
fix(home): restore poster and catalog skeletons with correct TV/mobile title visibility

Himanth-reddy:
Restores clean dynamic poster and catalog skeletons in the home feed. Reverts data-layer workarounds and implements conditional title rendering for TV poster lists where card titles are disabled.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/components/MediaCard.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/components/MediaCard.kt
@@ -98,6 +98,7 @@ fun MediaCard(
         PlaceholderCard(
             width = width,
             isLandscape = isLandscape,
+            showTitle = showTitle,
             modifier = modifier
         )
         return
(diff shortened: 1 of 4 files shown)


ProdigyV21:
The PR removes the Smooth Scrolling row/action, but `tvGeneralRowsForSection("appearance")` still includes row id `36`. On TV/D-pad this can create an invisible/unhandled focus stop in Appearance settings.

Please either:
1. fully keep Smooth Scrolling: restore the row, Enter action, and make Home respect `smoothScrolling`; or
2. fully remove it: remove row `36` from `tvGeneralRowsForSection`, remove the unused callback/state plumbing from SettingsScreen, and make the PR description clear that Smooth Scrolling is being removed.

After that, I think this is okay.

ProdigyV21:
Why does smooth scrolling need to be removed?

Himanth-reddy:
I originally mistakenly thought that the "Smooth Scrolling" toggle was doing nothing because even with the toggle turned off, the list still scrolled smoothly. But after checking it was working so i reverted the commits related to smooth scrolling

Himanth-reddy:
Sorry for making you confused

PR #445, opened 2026-07-07
web: Remove clock display from top navigation

Himanth-reddy:
## Summary

Remove the clock/time display from the web application's top navigation bar.

## Rationale

A clock display is unnecessary in a web application. Unlike a TV or set-top box interface where the app is the primary full-screen experience, web users always have convenient access to the current time through:

- **Operating system taskbar/menu bar** (Windows, macOS, Linux)
- **Browser tab bar or status area**
- **Mobile device status bar** (iOS, Android)

Displaying a redundant clock in the navigation consumes valuable UI space and adds unnecessary JavaScript timer overhead (interval polling every 30 seconds).

## Changes

- **`web/components/shell/TopNav.tsx`** — Removed clock state, `useEffect` interval timer, and the `<span className="top-clock">` element
- **`web/app/globals.css`** — Removed `.top-clock` style block and cleaned up responsive media query selectors that referenced it

### DIFF
--- a/web/app/globals.css
+++ b/web/app/globals.css
@@ -212,12 +212,6 @@ button {
   transform: scale(1.045);
 }
 
-.top-clock {
-  font-size: clamp(30px, 2.4vw, 46px);
-  font-weight: 350;
-  color: rgba(255, 255, 255, 0.86);
-}
-
 .content {
   min-height: 100vh;
   position: relative;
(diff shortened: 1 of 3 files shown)


ProdigyV21:
Thanks for this — the diff is clean, it typechecks, and there are no orphaned imports (the scroll listener still uses useState/useEffect). One small follow-up before it's fully tidy:

Removing the clock leaves the "Clock format" (24h/12h) selector in Settings now controlling nothing in the web UI — the clock was its only consumer. Could you also remove that now-dead row?

PR #446, opened 2026-07-08
Find best match (subtitles)

silentbil:
## Auto Find Best Subtitle Match                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
  Automatically selects the best-synced subtitle from your addons instead of a                                                                                                                                                                                                                                      
  blind release-name guess. It compares each candidate's cue timing against the                                                                                                                                                                                                                                     
  video's built-in subtitle track (or, when there's none, AI audio transcription)                                                                                                                                                                                                                                   
  and picks the one that actually lines up — falling back to AI translation or the                                                                                                                                                                                                                                  
  top release-name match if nothing is well-synced.                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
  ### Highlights                                                                                                                                                                                                                                                                                                    
  - **Timing-based matching** — no API key required for the core scan; works off                                                                                                                                                                                                                                    
    the built-in reference track.                                                                                                                                                                                                                                                                                   
  - **Per-stream cache** — a confirmed match is remembered per file (by                                                                                                                                                                                                                                             
    infoHash/videoHash/filename), so replays are instant. Keyed to the exact file,                                                                                                                                                                                                                                  
    not the title, so a different source re-scans.                                                                                                                                                                                                                                                                  
  - **Lives under Settings → Subtitles**, independent of the AI translation                                                                                                                                                                                                                                         
    feature. **Off by default** — opt in per taste.                                                                                                                                                                                                                                                                 
  - **Manual "Find Best Match"** entry in the player subtitle menu (forces a fresh                                                                                                                                                                                                                                  
    scan, ignoring the cache).                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                    
  ### Also included                                                                                                                                                                                                                                                                                                 
  - AIOStreams support: keep the addon's own stream order, parallel subtitle fetch                                                                                                                                                                                                                                  
    with retry (slow/502 endpoints), and correct VTT handling for `…/sub.vtt/` URLs.


ProdigyV21:
This is a really strong feature and I like the direction. I tested the merge locally: it merges cleanly, `git diff --check` passes, and both Play/Sideload Kotlin compiles pass.

One question, not sure if its intended or not. Any running Find Best Match scans when a new media item starts loading. `findBestSubtitleMatch()` captures the previous subtitle and can keep running for a while, but `loadMedia()` currently cancels subtitle/stream jobs without cancelling `findMatchJob`. If the user leaves the current title while a scan is active, the old scan can later restore/select a stale subtitle on the new playback session.

silentbil:
@ProdigyV21 good catch! 
fixed

PR #451, opened 2026-07-11
dolby Vision to HDR10 playback, audio passthrough, and full AIOStreams sources

silentbil:
- **Dolby Vision → HDR10**: DV profile-7 remux MKVs black-screened on devices without a                                                                                                                                                                                                                           
    DV decoder (audio played, video never rendered, source-failover churn). A vendored                                                                                                                                                                                                                              
    media3 1.9.0 Matroska extractor now strips the DV RPU/enhancement-layer NALs on the fly                                                                                                                                                                                                                         
    and rewrites the track to plain HEVC, so the HDR10 base layer plays. Device policy keeps                                                                                                                                                                                                                        
    native-DV devices (Shield-class) untouched; toggle in Settings → Playback (default on).                                                                                                                                                                                                                         
  - **Audio passthrough**: audio renderers now always prefer the platform MediaCodec path                                                                                                                                                                                                                           
    (FFmpeg stays fallback-only), so TrueHD/Atmos, DD+ and DTS-HD bitstream to the AVR                                                                                                                                                                                                                              
    instead of decoding to PCM — the FFmpeg-first Amlogic workaround now applies to video only.                                                                                                                                                                                                                     
  - **AIOStreams**: the source menu keeps the aggregator's server-side order, and dedup is                                                                                                                                                                                                                          
    scoped per addon — entries shared with standalone Torrentio/Debridio are no longer                                                                                                                                                                                                                              
    silently dropped (was 12 of 38 shown).


ProdigyV21:
Before merging, please fix these points:
- The custom Dolby extractor currently processes every HEVC MKV, including normal non-Dolby files. This adds full-frame buffering and allocations that could cause stuttering on 4K remuxes. Please only use the transformation path for confirmed Dolby Vision Profile 7 tracks.

- Profile 5 samples are preserved, but their track is still relabeled as normal HEVC. Profile 5 has no HDR10 base layer, so this could cause broken colours or failed playback. Please limit the format rewrite to supported profiles, currently Profile 7.

- Please add the Dolby Vision Compatibility toggle to mobile Playback & Controls settings too, since it is enabled by default.

After those changes and real-device testing with DV7, normal HEVC, Profile 5 and AVR passthrough, I think this will be a very strong addition.

PR #456, opened 2026-07-15
Improve find best match logic

silentbil:
Find best match feature, will now also try to correct subtitles with delay offset.

Deterministic built-in vs external: external subs stamped arvio-addon-sub:; absence of the stamp = genuinely muxed. No fuzzy matchin

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerScreen.kt
@@ -1269,33 +1269,42 @@ fun PlayerScreen(
                                 for (i in 0 until group.length) {
                                     val format = group.getTrackFormat(i)
                                     val formatTrackId = format.id?.trim().orEmpty()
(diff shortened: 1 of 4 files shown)


ProdigyV21:
Can you check this out?

- Reference intervals not overlapped by any candidate are removed before scoring. This can turn a subtitle matching only 2 of 8 cues into a 100% match.
- The offset search keeps the first equal-scoring result while scanning from `-10s`. In a targeted test where the correct offset was `+1.0s`, it selected `-9.5s`.
- The expensive offset scan runs on the main dispatcher for up to 10 candidates, which could cause playback/UI freezes on slower Android TV devices.
- If downloading a remembered offset subtitle fails, the original unshifted subtitle is selected while the UI still reports that the offset was applied.
- The PR also restricts AI translation to embedded English subtitles only. This removes support for external English and other embedded source tracks, which should be confirmed as intentional.

silentbil:
Thanks @ProdigyV21 
all Fixed

PR #464, opened 2026-07-19
fix cloud sync conflicts

silentbil:
Fix: multi-device settings sync reverting each other                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                    
  Problem                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                    
  Cloud settings are stored as one whole-account JSON blob written last-writer-wins with no clock. Any device that pushed — for any reason (playback progress, token refresh, or a pull's pre-push) — rewrote the entire blob with its own stale copy, silently reverting another device's setting change. No device
  had to even open settings; just being open and pushing was enough.                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                    
  Fix                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                    
  Per-field timestamp last-writer-wins for scalar settings, client-only (no backend change).                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  Android (CloudSyncRepository.kt)                                                                                                                                                                                                                                                                                  
  - Every scalar setting carries a fieldUpdatedAt timestamp (g:accentColor, p:<profileId>:autoPlayNext), stamped by diffing against a baseline — no per-write hooks.                                                                                                                                                
  - Push merges against the already-loaded remote: only overwrites a field it changed more recently → a stale device can't revert a peer.                                                                                                                                                                           
  - Apply merges the other way: adopts remote-newer fields, preserves local newer-unpushed ones → a pull can't lose your own just-made change.                                                                                                                                                                      
  - Applying a value stamps the winner's timestamp (not now()) → converges, no ping-pong. Skipped when the remote predates the feature, so rollout matches today until all devices update.                                                                                                                          
  - Scope: genuine globals + per-profile settings; list domains keep existing merges; defaultSubtitle keeps its own timestamp logic.                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                    
  Web (web/lib/cloud.ts, web/lib/store.tsx)                                                                                                                                                                                                                                                                         
  - saveCloudSettings now diffs against the session's last-synced baseline and writes/timestamps only the fields the web actually changed (same key scheme), instead of overwriting the whole settings blob. Web and phones no longer revert each other.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/CloudSyncRepository.kt
@@ -100,6 +100,12 @@ class CloudSyncRepository @Inject constructor(
     private val cloudSyncLocalDirtyAtKey = longPreferencesKey("cloud_sync_local_dirty_at")
     private val cloudSyncLastPushAtKey = longPreferencesKey("cloud_sync_last_push_at")
     private val cloudSyncLastAppliedAtKey = longPreferencesKey("cloud_sync_last_applied_at")
+    // Per-field last-writer-wins support (multi-device conflict resolution). `fieldTs` maps a
+    // canonical field key ("g:accentColor", "p:<profileId>:autoPlayNext") → epochMs of the last
+    // LOCAL change; `fieldBase` stores the value we last stamped, so we can detect local changes by
+    // diffing without hooking every write site.
+    private val cloudSyncFieldTsKey = stringPreferencesKey("cloud_sync_field_ts")
+    private val cloudSyncFieldBaseKey = stringPreferencesKey("cloud_sync_field_base")
     private val globalDnsProviderKey = stringPreferencesKey(OkHttpProvider.DNS_PROVIDER_PREF_KEY)
     private val customUserAgentKey = stringPreferencesKey(OkHttpProvider.USER_AGENT_PREF_KEY)
     @Volatile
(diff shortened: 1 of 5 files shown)


ProdigyV21:
Thanks for working on this. The overall per-field timestamp direction is good, and both Android tests and the web production build pass. However, these issues need fixing before it is safe to merge:

First-upgrade timestamp initialization
stampAndLoadFieldTs() timestamps every field with now() when no baseline exists. If another updated device already pushed newer cloud settings, the newly upgraded device gives all its stale local values later timestamps and reverts the cloud. Untracked fields must not be considered locally changed. Seed their baseline from the incoming remote payload unless there is evidence of an actual pending local change.

Web settings autosave can delete add-ons
saveCloudSettings() still writes addonsByProfile using the web session’s in-memory add-on list. Combined with Android’s new cloud-authoritative reconciliation, a stale web tab can remove add-ons installed on Android. Settings autosave must not mutate add-ons; only saveCloudAddons() with explicit add/remove operations should do so.

Canonical global settings are not read by web
pullCloudPayload() still reads global settings primarily from the legacy root.settings blob instead of timestamp-managed root fields such as accentColor, oledBlackBackground, customUserAgent, and AI subtitle settings. Android changes can therefore remain invisible on web. Map the canonical root fields into AppSettings and stop treating the stale whole-settings blob as authoritative.

Deleting the final add-on cannot propagate
reconcileAddonsWithCloud() treats every empty cloud list as invalid and keeps local add-ons. This means intentionally removing all add-ons never synchronizes. Add an add-on revision/tombstone or explicit removal metadata so an intentional empty state can be distinguished from a failed or incomplete pull.

PR #466, opened 2026-07-20
feat(plugin): support Nuvio JS scrapers and redesign settings UI

### LINKED ISSUE #459: Nuvio JS scraper repository installs but returns no sources
## Environment

- ARVIO: v1.9.981, original/unmodified Sideload build
- Device: Android emulator (LDPlayer 9)
- Reproduces outside any Cockpit/custom build.

## Summary

A valid Nuvio JS scraper repository can be installed and is shown as an HTTP local scraper bundle, but it returns no sources for movies or series. The same repository works in current NuvioTV.

Affected repository:

https://raw.githubusercontent.com/patr0nq/nuvioaddons/refs/heads/main/manifest.json

## Steps to reproduce

1. Install the repository URL in ARVIO from Addons or Plugins & Extensions.
2. ARVIO accepts it and shows a bundle with 13 HTTP providers. A brief "no catalogs" message may appear because this is a scraper repository rather than a catalog addon.
3. Open a movie or series and request sources.
4. No sources are returned.
5. Install the same repository in current NuvioTV and repeat the test; sources are returned there.

## Expected behavior

Nuvio JS repositories should execute through the JS plugin runtime and return their streams. If a repository cannot be handled by the HTTP-local fallback, ARVIO should not install it as a working HTTP provider bundle that can only return an empty result.

## Controls

This raw Nuvio JS repository works in ARVIO, so raw GitHub manifests and the general repository format are supported:

https://raw.githubusercontent.com/saimuelbr/saimuel-nuvio-repo/refs/heads/main/manifest.json

Standard Stremio addons also work normally, for example:

https://hdhub.thevolecitor.qzz.io/eyJ0b3Jib3giOiJ1bnNldCIsInF1YWxpdGllcyI6IjIxNjBwLDEwODBwLDcyMHAiLCJzb3J0IjoiZGVzYyJ9/manifest.json

This narrows the issue to compatibility with the more advanced patr0n provider scripts rather than the manifest URL or Cockpit integration.

### HTTP-local fallback

`HttpLocalScraperRuntime.fetchInstallCandidate()` accepts enabled entries whose formats include HTTP media formats. However, `resolveHttpStreams()` dispatches work only for a fixed set of provider IDs such as `multivid`, `videasy`, `vidlink`, `vidsrc`, `rgshows`, `playimdb`, `dooflix`, `fmovies`, `brazucaplay`, and `netmirror`.

The affected manifest uses IDs such as `patron-film-makinesi`, `patron-dizipal`, and `patron-sinewix`. These entries can therefore be counted and displayed as HTTP providers while producing zero resolver jobs.

Suggested behavior: only claim a Nuvio manifest as an HTTP-local bundle when at least one provider ID has an implemented resolver, or route the manifest exclusively through the JS plugin runtime.

### JS runtime parity

The affected providers use multiple fetch chains, HTML/Cheerio parsing, redirects, `Promise.all`, `AbortSignal`, base64 decoding, and potentially large HTML responses.

Compared with current NuvioTV, ARVIO v1.9.981 has notable runtime differences:

- ARVIO limits fetched response bytes/body characters to 256 KiB; current NuvioTV uses 1 MiB.
- ARVIO exposes `__native_fetch` via `asyncFunction` and a suspending OkHttp bridge; current NuvioTV exposes a synchronous native function to QuickJS.
- Current NuvioTV uses IPv4-first DNS for the plugin HTTP client.

Suggested investigation: align ARVIO's Sideload `PluginRuntime` fetch bridge and response limits with current NuvioTV, then verify the affected repository again.

## Acceptance criteria

- The affected repository no longer appears as a functional HTTP bundle when none of its provider IDs have HTTP-local resolvers.
- Its enabled JS scrapers execute through PluginRuntime.
- At least one supported movie and series returns sources as it does in current NuvioTV.
- Existing working Nuvio repositories and standard Stremio addons remain unaffected.

Himanth-reddy:
This PR fixes compatibility for Nuvio JS scraper plugins and redesigns the Plugins settings UI to align with other screens.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -122,7 +122,6 @@ android {
 
             // Build config fields for release
             buildConfigField("Boolean", "ENABLE_CRASH_REPORTING", "true")
-            buildConfigField("Boolean", "FEATURE_PLUGINS_ENABLED", "false")
         }
 
         debug {
@@ -147,7 +146,6 @@ android {
             isJniDebuggable = false
 
             buildConfigField("Boolean", "ENABLE_CRASH_REPORTING", "true")
-            buildConfigField("Boolean", "FEATURE_PLUGINS_ENABLED", "false")
         }
     }
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -122,6 +122,7 @@ android {
 
             // Build config fields for release
             buildConfigField("Boolean", "ENABLE_CRASH_REPORTING", "true")
+            buildConfigField("Boolean", "FEATURE_PLUGINS_ENABLED", "false")
         }
 
         debug {
@@ -146,6 +147,7 @@ android {
             isJniDebuggable = false
 
             buildConfigField("Boolean", "ENABLE_CRASH_REPORTING", "true")
+            buildConfigField("Boolean", "FEATURE_PLUGINS_ENABLED", "false")
         }
     }
 
--- a/app/src/main/kotlin/com/arflix/tv/data/repository/HttpLocalScraperRuntime.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/repository/HttpLocalScraperRuntime.kt
@@ -63,7 +63,7 @@ class HttpLocalScraperRuntime @Inject constructor(
     ): HttpLocalScraperInstallCandidate? = withContext(Dispatchers.IO) {
         val manifestUrl = manifestUrlFor(url)
         val manifest = fetchManifest(manifestUrl) ?: return@withContext null
-        val httpScrapers = manifest.scrapers.filter { it.isHttpOnlyEnabled() }
+        val httpScrapers = manifest.scrapers.filter { it.isSupportedHttpLocalScraper() }
         if (httpScrapers.isEmpty()) return@withContext null
 
         val stableId = "http.local.${shortHash(manifestUrl)}"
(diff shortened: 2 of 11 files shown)


ProdigyV21:
Is this finished or you still working on it? I looked at it and these are my findings so far.

Findings
[P1] It conflicts with current main. The conflict is in SettingsScreen.kt between catalog-pack modal handling and the new plugin modal handling. Rebase and preserve both conditions.

[P1] Plugins are disabled in production sideload builds. [app/build.gradle.kts:125](https://github.com/ProdigyV21/ARVIO/blob/b5e6b61e9a5fec07b0c55fa844ca7b8cfe2b97a9/app/build.gradle.kts#L125) overrides FEATURE_PLUGINS_ENABLED to false for release and staging. I confirmed:
Sideload Debug: true
Sideload Release/Staging: false
Remove those build-type overrides and retain the existing flavor values: Play false, Sideload true.

[P2] Mobile opens the wrong settings page. [SettingsScreen.kt:3900](https://github.com/ProdigyV21/ARVIO/blob/b5e6b61e9a5fec07b0c55fa844ca7b8cfe2b97a9/app/src/main/kotlin/com/arflix/tv/ui/screens/settings/SettingsScreen.kt#L3900) routes both “Addons” and “Plugins & Extensions” to the Stremio addons screen. Mobile cannot access the redesigned plugin screen.

[P2] Reset does not completely unload extensions. [ExternalExtensionLoader.kt:661](https://github.com/ProdigyV21/ARVIO/blob/b5e6b61e9a5fec07b0c55fa844ca7b8cfe2b97a9/app/src/sideload/kotlin/com/arflix/tv/core/plugin/cloudstream/ExternalExtensionLoader.kt#L661) deletes files and caches, but extension extractors already registered globally can remain active until the app restarts.

[P2] Reset can leave TV focus out of range. After clearing a long plugin list, the selected index is not reset or clamped, temporarily leaving no visible focused row.

Himanth-reddy:
I thought i would get some response from the testers but they didn't so let them complain in the next update.

PR #468, opened 2026-07-20
remember played video timestamp

silentbil:
Details screen: keep season/episode context                                                                                                                                                                                       
                                                                                                                                                                                                                                    
  Fixes two UX issues on the series details screen:                                                                                                                                                                                 
                                                                                                                                                                                                                                    
  - Season list didn't reflect episodes on hover. Moving focus between seasons now loads that season's episodes automatically (100ms debounce so quickly stepping 1→4 only fetches the season you land on).                         
  - Returning from the player reset to S1E1. On a loaded device the details screen is recreated behind the player, and loadDetails was snapping back to the stale nav-arg target instead of where the user actually was. On re-entry
  it now:                                                                                                                                                                                                                           
    - keeps the season the user was browsing (from the surviving ViewModel state), and                                                                                                                                              
    - points the Continue button at the episode they last played (recorded even for brief <60s plays, which don't create a resume point), falling back to resume → next-unwatched.


ProdigyV21:
Review findings:
Season requests can overlap. The 100 ms delay only cancels before loading starts. Once started, season loads run separately. An older request could finish last and display Season 2’s episodes while Season 3 is selected. Clicking a season can also start the same request twice because the click handler and new effect both call loadSeason().

Episode focus is not reset during automatic switching. If Episode 8 was selected and you move to another season, moving down can still focus Episode 8 instead of Episode 1. The existing click-based season switch resets this correctly, but the new automatic path does not.

When this is fixed I can merge. Good addition.

silentbil:
thanks @ProdigyV21 
deployed

PR #488, opened 2026-07-24
Mdblist support

silentbil:
Support mdblist as well as trakt

- android tv 
- mobile
- web


ProdigyV21:
nice work, this is a clean abstraction. The shared SyncClient interface with both providers returning Trakt-shaped data is exactly right: all existing mappers keep working unchanged, Trakt tokens stay untouched, and the mutual exclusion means existing Trakt users notice nothing. Also checked the proxy route for SSRF — host is hardcoded to api.mdblist.com and new URL() keeps it there, so that's solid.

Two things I'd like fixed before merging, both around serverless cost (context: we had a Netlify blowup earlier this year — ~500 credits/day — and moved TMDB/Trakt/catalog traffic to a Cloudflare Worker to stop it, so I'm cautious about anything that multiplies function invocations):

1. fetchAllWatched() runs twice per refresh

watched("movies") and watched("shows") each call fetchAllWatched() in full, and store.tsx calls both in the same Promise.all. So every refresh paginates the entire watched history twice — double the /api/mdblist invocations for zero benefit. Memoizing the result (or sharing the in-flight promise) with a short TTL fixes it:

private watchedCache: { at: number; data: Promise<...> } | null = null;

private fetchAllWatched() {
  if (this.watchedCache && Date.now() - this.watchedCache.at < 30_000) return this.watchedCache.data;
  const data = this.loadAllWatched();       // the current while-loop body
  this.watchedCache = { at: Date.now(), data };
  return data;
}

2. while (true) has no page cap

The pagination loop only exits on an error or has_more: false. If MDBList ever returns has_more: true with empty pages, that's an infinite loop of serverless invocations. Low probability, high blast radius — please add a hard cap:

for (let page = 0; page < 20; page += 1) { ... }

Minor: the proxy accepts the key as an ?apikey= query param as well as the x-mdblist-key header. Query params end up in CDN/server logs — header-only would be cleaner.

Not blocking, just flagging for later: /api/mdblist is a Next.js route handler, so it's a Netlify function per call. It's opt-in and low-volume, so it's fine for now — but if MDBList gets popular this is the first thing we should move to the Cloudflare Worker (same as /api/tmdb).

Happy to merge once 1 and 2 are in. Thanks for doing all three platforms in one go 🙌

silentbil:
@ProdigyV21 fixed

PR #495, opened 2026-07-28
feat(web): add AniList and TVDB metadata with BYOK

Himanth-reddy:
## Summary
Adds AniList and TVDB v4 metadata support to the ARVIO web app, with provider fallback chains and Bring Your Own Key settings. The original feature implementation remains credited to @Himanth-reddy.

### What works
- Anime metadata: AniList, then TVDB when configured, with TMDB as the canonical fallback.
- TV metadata: TVDB when configured, with TMDB as the canonical fallback.
- Movies remain on TMDB.
- Custom TMDB and TVDB credentials stay local to the browser and are excluded from cloud settings.
- Provider IDs are mapped before dispatch, so TMDB, TVDB, and AniList IDs are never mixed.
- TMDB movie/TV identity and season/episode numbering remain canonical for details, playback, tracking, and addons.
- TVDB can enrich matching normal-TV episode metadata without replacing playback numbering.

### Scope
This is a complete web vertical slice. The inactive Android scaffolding was removed; an Android implementation can follow separately once its settings and runtime consumers are ready.

### Verification
- `npm run build`
- `:app:testPlayDebugUnitTest`
- `:app:testSideloadDebugUnitTest`
- `git diff origin/main --check`


ProdigyV21 (changes requested):
Thanks Himanth — the direction here is right (AniList for anime, TVDB as an opt-in BYOK provider, priority fallback), the per-file code is clean, and both builds pass. TVDB being hard-disabled without a user key is exactly the right call.

The blocker: nothing is wired up yet, so merging this would ship a settings page for a feature that doesn't exist. Concretely:

1. **No call sites.** `MetadataDispatcher` (web and Android), the resolvers, and the AniZip mapper are never referenced outside the new files — this PR adds no consumers either. Details, search and episodes still go 100% through the existing TMDB path on both platforms, so the ~840 added lines are unreachable at runtime.

2. **The anime chain can't trigger.** `MediaType` is `"movie" | "tv"` (`web/lib/types.ts`) and nothing anywhere produces or detects `"anime"` — the `"anime" as any` casts paper over this. The feature needs a real anime media type or detection (e.g. TMDB genre 16 + origin country JP, or an AniList-id source) before a priority chain for it means anything.

3. **The dispatcher's terminal fallback doesn't exist.** `dispatcher.ts` only registers `anilist` and `tvdb` in its resolver map — there is no TMDB resolver. As-is, movies (chain `["tmdb"]`) resolve to `null` for everyone, and TV without a TVDB key also dead-ends. Wiring the dispatcher into the app in this state would break details for all non-BYOK users.

4. **`customTmdbApiKey` is stored but never read.** The real TMDB fetch path (`web/lib/tmdb.ts` → `/api/tmdb` proxy) doesn't consult it, so the BYOK TMDB field does nothing.

5. **The settings page overstates reality.** "Active chain: AniList → TVDB → TMDB" reads as live status, but nothing consults `metadataAnimeProviders` / `metadataTvProviders` / `metadataMovieProviders`. A user can enter a key and see "Active — TVDB enabled" while zero requests ever hit TVDB.

Smaller points for the next revision:

- The web TVDB token cache (`cachedToken` in `tvdb.ts`) isn't invalidated when the key/PIN changes — log in with key A, switch to key B, and requests keep riding A's token for up to 23h.
- Android's `getAnimeDetails(query)` searches by name only and ignores the priority config — the Kotlin `MetadataDispatcher` has no actual fallback chain despite the name.
- The settings UI uses inline `style={{…}}` throughout; the rest of the app styles via classes in `globals.css`.
- `convertAniListToTmdbEpisode` is exactly what playback needs for absolute-numbered anime episodes — it just needs a call site in the episode/stream resolution flow to earn its place.

**What would make this mergeable:** land one vertical slice end-to-end instead of scaffolding everywhere. Suggested shape, web-only first:

1. Add anime detection (or an explicit anime type) for titles.
2. Register a TMDB resolver as the terminal fallback so today's behavior is the guaranteed worst case.
3. Route `getDetails`/`getEpisodes` for detected anime through the dispatcher.
4. Use the AniZip mapping when resolving episode streams (that's where it pays off).
5. Only then surface the settings UI for keys/priorities, reflecting what's actually consulted.

Android can follow the same shape once the web slice proves out. Happy to review that version quickly.

Himanth-reddy:
Thanks for the detailed review! All 5 blockers and the smaller points have been addressed in commit `6dffa928`. Here's what was done:

### Blockers addressed

1. **Call sites wired up.** `getDetails` and `getSeasonEpisodes` in `tmdb.ts` now route through `MetadataDispatcher` when a priority config is active. `store.tsx` constructs a `ProviderPriorityConfig` from the user's settings and passes it into every details fetch. The dispatcher code is now reachable at runtime.

2. **Anime detection implemented.** Added `isAnime()` helper in `tmdb.ts` (checks TMDB genre ID 16 + Japanese origin country or original language `ja`). Extended `MediaType` to `"movie" | "tv" | "anime"` — removed all `as any` casts in `anilist.ts`.

3. **TMDB resolver registered as terminal fallback.** Created `web/lib/metadata/tmdbResolver.ts` implementing the full `MetadataResolver` interface (details, episodes, search). Registered it in `dispatcher.ts` so every priority chain guaranteed falls back to TMDB — non-BYOK users are never broken.

4. **`customTmdbApiKey` is now consumed.** The `tmdb<T>()` fetch helper in `tmdb.ts` appends `customTmdbApiKey` when provided via `ProviderPriorityConfig`. The TMDB resolver also passes it through.

5. **Settings UI reflects reality.** The settings section now only displays provider chains that are actually consulted by the dispatcher. Priorities are read from `metadataMovieProviders` / `metadataTvProviders` / `metadataAnimeProviders` which are passed directly into the fetch flow.

### Smaller points fixed

- **TVDB token cache invalidation:** `tvdb.ts` now stores `cachedKey` and `cachedPin` alongside the token. When either changes, the token is immediately invalidated — no more riding a stale token for 23h.
- **Settings UI inline styles removed:** `MetadataSection` in `SettingsScreen.tsx` now uses standard CSS classes (`set-row`, `set-label`, `set-control`, `settings-input`) consistent with the rest of the app.
- **AniZip mapping wired into episode streams:** `store.tsx` now calls `fetchAniZipMapping` + `convertAniListToTmdbEpisode` in `loadEpisodeStreams` for anime items to convert absolute episode numbers to TMDB/TVDB season+episode before calling stream providers.
- **Type safety cleanup:** Propagated the expanded `MediaType` union across `sync.ts`, `trakt.ts`, `mdblist.ts`, and `imdbRatings.ts`. Fixed type predicate widening issues with `NonNullable` generics.

### Verification
- `npm run build` passes cleanly (✓ compiled, ✓ types, ✓ static pages).

Android vertical slice will follow once this web slice is validated. Ready for re-review!

ProdigyV21:
I reviewed the latest PR head against current `main`. The exact merge is conflict-free, and the web production build, both Android Kotlin compilations, and both Android unit-test suites pass. However, there are runtime blockers that the current tests do not cover:

1. **[P1] Anime details and playback are broken** (`web/lib/tmdb.ts:109`). The PR changes anime from `tv` to a new `anime` media type, but existing code still only recognizes `tv`. This leads to the invalid TMDB endpoint `anime/{id}`, hides seasons, queries addons as movies, omits episode numbers, and disables next-episode playback. Please either keep the canonical type as `tv` with a separate anime flag, or update every TMDB/UI/addon/playback boundary consistently.

2. **[P1] TMDB IDs are incorrectly used as AniList/TVDB IDs** (`web/lib/tmdb.ts:1111`). `item.id` comes from TMDB, but AniList treats it as an AniList ID and TVDB treats it as a TVDB ID. These ID namespaces are unrelated, so a matching number can return a completely different title and overwrite the correct metadata. Map external IDs first and preserve the canonical TMDB ID separately.

3. **[P1] Movie metadata can be replaced by an unrelated TV show** (`web/lib/metadata/tmdbResolver.ts:14`). The TMDB resolver always tries `tv/{id}` before `movie/{id}`. If the same numeric ID exists in both namespaces, opening a movie can display TV-show metadata. Pass the requested media type into the resolver and query only the correct endpoint.

4. **[P2] AniZip episode conversion receives the wrong ID** (`web/lib/store.tsx:1299`). It sends the TMDB ID as `anilist_id`, so the mapping normally fails or may resolve incorrectly. Obtain the actual AniList ID before calling AniZip.

5. **[P2] TVDB episode loading only reads page zero** (`web/lib/metadata/tvdb.ts:87`). Later seasons of long-running shows can be missing. Follow TVDB pagination until all pages are loaded.

6. **[P2] BYOK credentials use the ordinary cloud-settings path** (`web/lib/types.ts:463`). The keys are included in `AppSettings`, which is cloud-synced as a plaintext settings object, and the TMDB key is also placed in a URL query string. Please make the security/storage behavior explicit and avoid exposing secrets in URLs.

7. **[P2] The Android integration is currently inactive scaffolding** (`app/src/main/kotlin/com/arflix/tv/data/repository/MetadataDispatcher.kt:15`). The APIs and dispatcher are registered, but there are no Android consumers or settings wiring, so the feature does not yet work on Android.

The feature direction is useful, but I do not recommend merging this version yet because it can break existing anime playback and show incorrect metadata for ordinary movies. Please also add focused tests for provider-ID mapping, movie/TV namespace collisions, anime details/streams, and AniZip conversion. `git diff --check` currently reports a few minor whitespace errors as well.

PR #501, opened 2026-07-30
feat(web): add media context menus and persistent watched actions

Himanth-reddy:
## Summary
- Adds desktop right-click and mobile long-press context menus for media cards, episodes, and seasons.
- Supports watchlist, watched/unwatched, Continue Watching dismissal, episode playback, and season watched actions.
- Adds responsive, accessible context-menu styling for desktop and mobile.

## Review fixes
- Rebased onto current `main` and kept both the current no-addons prompt and the new context menu.
- Prevents a completed long-press from also opening Details.
- Closes every custom menu action before it runs.
- Removes the duplicate Play action that only opened Details.
- Persists watched changes using the correct 0-1 cloud progress scale.
- Removes cloud resume data and Trakt playback/progress without deleting watched history.
- Loads the season that was actually clicked and persists season actions through cloud and the active sync provider.
- Avoids reporting watchlist success when no Trakt/MDBList provider is connected.

## Validation
- `npm run build` passes against the exact merge with current `main`.
- `git diff --check` passes.


ProdigyV21 (dismissed):
Thanks Himanth — there's real value here: the right-click/long-press context menu framework (pointer-only, so TV nav is untouched), the IPTV sort-order setting, and the skeleton/shimmer polish are all genuine improvements.

But this bundles 4+ features, and several collide with what landed on main this week:

1. **Watched/Continue-Watching sync corrupts data.** `toggleWatched` passes `progress: 1` to `saveProgress`, but the cloud payload uses a 0–100 scale — cloud.ts only drops an item from Continue Watching at `progress >= 90`, otherwise it inserts it at the TOP. "Mark as watched" therefore pushes the title into Continue Watching at 1% on every device, including the Android app. `removeFromContinueWatching` also never touches the cloud payload, so removed items resurface on the next refresh.

2. **The Live TV padding hunk re-breaks a landed fix.** Main fixed the top-bar overlap at a measured `clamp(84px, 9vh, 108px)` (e153ff58); this PR's 112–154px predates that and adds ~40px of dead space. Drop the hunk.

3. **The always-mounted AppShell is too costly on TV browsers** (our primary target): all six screens mount at boot and hidden screens keep working — the hero keeps rotating and fetching logos/ratings forever. The `display: contents` wrappers also break the `.content > .screen` rule.

4. **The hero rework regresses main's behavior**: previous title's logo shows during swaps (main synchronizes those on purpose), hover/focus-to-pin is removed, and the 12 dots become D-pad focus stops.

5. **Please remove**: `version.json` (build-generated, never commit), the TMDB host swap, the unused plugin/scraper types and settings fields, and the logo→PNG swaps.

Could you split it? PR 1: context menu with fixed watched/CW sync (+ `role="dialog"` on the scrim, CSS for the header-art/footer classes). PR 2: IPTV sort order — nearly mergeable as-is. PR 3: skeletons/shimmer. I'd gladly take all three.

Himanth-reddy:
Thanks for the detailed review!

I've updated this PR to focus specifically on **Context Menu & Watched/CW Sync**:
1. **Fixed Watched & Continue Watching Sync**: `toggleWatched` now passes `progress: 100` so `cloud.ts` recognizes completion (`progress >= 90`) and drops the item from Continue Watching instead of placing it at 1%. `removeFromContinueWatching` also updates the cloud payload to prevent removed items from resurfacing.
2. **Context Menu Accessibility & CSS**: Added `role="dialog"` and `aria-modal="true"` to the context menu scrim, along with `.context-menu-header-art` and `.context-menu-footer-hint` CSS classes.
3. **Reverted Unrelated Features**: Reverted the always-mounted AppShell, standalone hero carousel rework, Live TV padding change, version.json, TMDB proxy change, and unused types/settings fields.

Regarding point 3 (AppShell always-mounted for TV browsers) — why would people use the web app in a TV's browser if there is already the native app?

ProdigyV21:
Thanks for cleaning up the scope. The context-menu feature is a useful addition, and the current branch merges cleanly into main. Both GitHub CI and a local merged web production build pass.

There are still a few behavior issues to fix before merging:

1. **Remove from Continue Watching** calls the Trakt/MDBList remove-from-history API. This can erase watched history, while the provider's paused-playback entry is not removed, so the item may return after refresh. Please use a dedicated dismissal/playback-removal path without changing watched history.
2. **Season watched/unwatched actions** iterate the episodes loaded for the currently selected season, even when a different season tab was right-clicked. They are also local-only and disappear after refresh. Load the clicked season's episodes and persist the result through cloud/Trakt/MDBList.
3. **Mobile long-press** can still trigger the card's normal click when the finger is released, opening Details behind the menu. Track when the long-press fires and suppress the following click.
4. **`saveProgress` expects progress from 0 to 1**, but the new actions pass `100`. Please use `1` for completed progress.
5. **Play and View Details currently do the same thing**: both call `openDetails(item)`. Please wire Play to the playback/source flow, or remove/rename the duplicate action.

Minor cleanup: remove the trailing whitespace in `AppShell.tsx` and update the PR title/description, since the carousel and always-mounted tab changes are no longer part of the final six-file diff.

Once these points are addressed, this should be in much better shape to merge.

ProdigyV21 (approved):
All requested changes are addressed. The exact merge passes the web production build, repository Build Check, patch checks, and the PR labeler.

PR #502, opened 2026-07-31
fix(mobile): resolve touch issues and unify button styling in plugins settings

Himanth-reddy:
### Summary
This PR fixes touch interaction issues on Android mobile devices within the Plugins settings screen and aligns button and category styling with the rest of the mobile settings sections.

### Changes
1. **Mobile Touch Layout in PluginScreen**:
   - Added a dedicated mobile layout branch using `LocalDeviceType.current.isTouchDevice()` so touch events trigger directly without TV D-Pad focus slot overrides.
   - Replaced TV-only `androidx.tv.material3.Surface` dialog buttons in `AddRepoDialog` with touch/focus-friendly Compose `Box` buttons.
2. **Unified Styling via MobileSettingsCategory and MobileSettingsRow**:
   - Shared `MobileSettingsCategory` and `MobileSettingsRow` in `SettingsRows.kt`.
   - Used standard mobile category cards (`BackgroundElevated`, `RoundedCornerShape(12.dp)`), standard row padding, dividers, value pills, and green toggle switches (`SuccessGreen`) for plugins settings.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/components/SettingsRows.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/components/SettingsRows.kt
@@ -29,7 +29,9 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.sp
 import androidx.tv.material3.ExperimentalTvMaterial3Api
 import com.arflix.tv.R
+import androidx.compose.ui.draw.clip
 import com.arflix.tv.ui.theme.ArflixTypography
+import com.arflix.tv.ui.theme.BackgroundElevated
 import com.arflix.tv.ui.theme.Pink
 import com.arflix.tv.ui.theme.SuccessGreen
 import com.arflix.tv.ui.theme.TextPrimary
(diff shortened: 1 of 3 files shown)


ProdigyV21 (changes requested):
Thanks — the core here is genuinely good. The mobile card layout for plugins matches the rest of mobile settings, the `MobileSettingsCategory`/`MobileSettingsRow` extraction to SettingsRows.kt is a clean exact move, and replacing the tv.material3 Surface dialog buttons fixes a real touch issue.

Three blockers, all on the TV side (TV is the primary form factor):

1. **Reset Plugins & Extensions is silently removed** — the reset row, `WarningDialog`, and the only UI path to `PluginUiEvent.ResetAllPlugins` are gone on both TV and mobile, leaving dead code behind. Please restore it in both branches. If you think it should be dropped, propose that separately.

2. **The new Box+clickable dialog buttons have no focused-state visual.** On TV you can no longer see whether Cancel or Add is focused. Keep the touch fix, but add a focus indicator (interactionSource/onFocusChanged driving background or border), or keep the TV Surface on the non-touch branch.

3. **Please revert removing `.focusProperties { canFocus = false }`** from the Add row and `FocusableSettingsRow`/`FocusableSettingsToggleRow`. Those are TV-only paths now (mobile takes your new branch), so the change buys nothing for touch — and that guard is what keeps real Compose focus out of rows in the virtual-focus system.

Also worth considering: mobile repo rows delete instantly on any tap — a confirmation would prevent accidents.

With 1–3 addressed, happy to merge.

ProdigyV21 (approved):
Reviewed and verified after updating to current main. The requested mobile touch fixes, restored reset flow, TV focus guards, and dialog focus states are present. Play and Sideload Kotlin compilation passed locally, the Sideload unit-test suite passed locally, and GitHub's required Verify build check passed.

PR #503, opened 2026-07-31
ci: add PR labeler and Actionlint automation

Himanth-reddy:
## Summary
- Adds automatic PR area labels based on changed paths.
- Adds Actionlint validation for GitHub workflow files.
- Stops ignoring `.github/workflows/` so new workflow files can be tracked.

## Review changes
- Renovate, Release Please, and Danger were removed as requested.
- Rebased onto current `main` while preserving the current build/signing workflow.
- Added the configured `area:*` labels to the repository without increasing workflow permissions.

## Validation
- All workflow files pass Actionlint locally.


ProdigyV21 (dismissed):
Thanks — the config quality is careful (labeler globs match our real layout, the release-please manifest matches the latest tag, and the SC2129 cleanup shows you ran actionlint against the repo). Actionlint and the labeler are welcome.

Three pieces can't go in:

1. **Renovate**: there's no `RENOVATE_TOKEN` secret, and the `GITHUB_TOKEN` fallback can't initialize self-hosted Renovate — the Saturday cron would fail every week. Even with a PAT, Renovate's PRs wouldn't trigger build-check, so dependency bumps (AGP/Compose/Next) would land with zero CI. The config also uses deprecated `matchPaths`/`matchPackagePatterns`.

2. **Release Please**: release-type `simple` only maintains version.txt/CHANGELOG — it can't bump `versionName` in app/build.gradle.kts (the real APK version), and web's version.json is a build-generated stamp that's never committed. Its first `feat` commit would also propose v1.10.0, breaking the 1.9.98x scheme. We deliberately removed release automation in 1be60644, and this opens a release PR on every push to main.

3. **Danger**: a summary comment on every PR is noise at two-person scale — drop it, or warns-only without the comment.

Mergeable version: remove renovate.yml + renovate.json, release-please.yml + both configs, and danger.yml + Dangerfile.js; keep actionlint, labeler, the build-check cleanup, and the .gitignore change. Happy to merge that.

ProdigyV21 (approved):
Requested changes are addressed, the branch is current with main, and Actionlint plus the Android build both pass.

PR #504, opened 2026-08-03
feat: "Only Iptv" mode

Aaronnn17:
For users who only access content hosted on an IPTV server, it can be confusing to see covers for content that hasn't yet been uploaded to the server. 

In this mode, ARVIO skips the other catalogs and loads only the one for content hosted on the server, then loads the covers and information later, so that only the content that can be viewed is visible.

This is an extra feature for those who use the app only as an IPTV player.

To enable it, I've added a button to the IPTV menu where you can turn this option on or off.


ProdigyV21 (changes requested):
Thanks Aaron — "IPTV-only mode" is a genuinely good idea; plenty of people use ARVIO purely as an IPTV player. But this version can't go in:

1. **It doesn't compile.** Missing comma before the new `onToggleIptvOnlyMode` parameter, the closing brace of `IptvSettings` is deleted, and the three `IptvSettings(...)` call sites never pass the two new required parameters. Verified locally: the build fails with an `Unresolved reference` cascade.

2. **Permanent data loss.** The filter sits inside `readCatalogsFromPrefs`, but `readCatalogsForActiveProfile` treats any difference from raw prefs as a migration and calls `saveCatalogs()`. The first home-screen load after enabling the toggle permanently rewrites the saved catalog list — destroying Trakt/MDBList/addon catalogs — and syncs the wipe to the cloud. Toggling off restores nothing.

3. **D-pad breakage.** The parent screen's `sectionMaxIndex` and OK-press handler still use the old row indices. With the shifted slots, OK on "Refresh" now runs `clearIptvConfig()` (deletes all playlists), the Delete row is unreachable, and the new toggle can't be activated by remote. TV is the primary form factor.

4. **Wrong target.** `HOME_SERVER` catalogs are Plex/Jellyfin/Emby, not IPTV — IPTV playlists live in Live TV and never pass through CatalogRepository. Enabling the mode keeps Plex rows and blanks home for actual IPTV-only users.

Happy to review a rework: filter at the presentation layer (HomeViewModel), never in the persistence read path; hide VOD rows and land on Live TV; update the D-pad indices together with the rows.

PR #505, opened 2026-08-03
fix: “Continue watching” does not display titles with the updated language

Aaronnn17:
The “Continue Watching” section displays titles from a local database, so if you change the language in the settings, it doesn't sync with the new language and continues to display them in the old language, while other sections like “Trending Shows/Movies” do display them in the updated language.

Added a minor fix so that the “Continue Watching” repository retrieves data in the updated language.

photo_2026-08-03_11-23-56


ProdigyV21 (changes requested):
Thanks Aaron — this targets a real bug: the launcher "Continue watching" channel keeps titles in whatever language was active when the row was written, and re-fetching through MediaRepository (which passes the user's content language to TMDB) is the right mechanism.

Three blockers:

1. **It doesn't compile.** The last hunk deletes the closing brace of `loadPublisherItems()` without replacing it. Verified locally: `Unresolved reference` cascade through the rest of the file and MainActivity. Please run a local build before pushing.

2. **The fetches run for every history entry.** The new network calls sit inside the `.map` over the full history list, before `.distinctBy`/`.take(50)`. A large history means hundreds of sequential TMDB calls — and `getSeasonEpisodes` is heavyweight (Trakt watched-cache init plus per-episode ratings). `refreshForCurrentProfile()` fires from many places, so: dedupe and `take(50)` first, then resolve titles for at most those 50.

3. **It only fixes the fallback path.** The primary path — `traktRepository.getContinueWatching()`, where most users' rows come from — still publishes English titles (no language param in its TMDB calls). Cover it too, or the bug still reproduces for Trakt users.

Minor: drop the redundant same-package `MediaRepository` import and the unnecessary safe-calls (`getTvDetails` returns non-null).

Happy to re-review once it builds and the fetch is capped.

Aaronnn17:
Closing this to open a new PR from the correct branch

Issue #420, opened 2026-06-30
Please add IPTV playlists to be provided order and not channel number order.


Collier14:
Could we please get the option to have IPTV in playlist providers order and not channel number order. 
Have had alot of people request this on Reddit and discord. 
Every IPTV app runs your groups and channels in providers order and not channel numbers. 
Thank you.

PR #510, opened 2026-08-04
feat(android): add IPTV Provider Sort Order setting option

Himanth-reddy:
### Summary
This PR adds the IPTV Provider Sort Order setting option for Android (Mobile & TV), bringing feature parity with the Web platform (resolving Issue #420).

### Changes
- Added `sortOrder` DataStore persistence (`"provider" | "number" | "name"`) in `IptvRepository.kt`.
- Exposed `iptvSortOrder` state and `setIptvSortOrder` in `SettingsViewModel.kt` & `TvViewModel.kt`.
- Added Sort Order toggle row in Android Settings (`SettingsScreen.kt`) for both Mobile and TV form factors.
- Applied category and channel sort mode handling in `LiveTvScreen.kt`.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/model/IptvModels.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/model/IptvModels.kt
@@ -79,6 +79,7 @@ data class IptvSnapshot(
     val favoriteChannels: List<String> = emptyList(),
     val hiddenGroups: List<String> = emptyList(),
     val groupOrder: List<String> = emptyList(),
+    val sortOrder: String = "provider",
     val epgWarning: String? = null,
     val loadedAt: Instant = Instant.now()
 )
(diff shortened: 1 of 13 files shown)


ProdigyV21:
Thanks, this is a useful addition and the exact merge into current `main` builds successfully for both Play and Sideload. A few things still need fixing before it is ready:

1. On TV, the new Sort order row is unreachable. It uses `playlists.size + 3`, but the IPTV section max remains `playlists.size + 2`, and the Enter handler has no case for the new row. Please update the max/focus order and Enter action.
2. Channel Number currently sorts by `EnrichedChannel.number`, which is ARVIO's generated row index and already follows provider order. Please sort by the numeric value of `source.providerChannelNumber` instead, with missing or invalid numbers kept stable or placed last.
3. The subtitle says channels and groups are sorted, but only channels are changed. Please apply the selected mode to playlist groups too, or change the description if groups are intentionally excluded.
4. `saveSortOrder()` marks IPTV cloud data dirty, but `IptvCloudProfileState` export/import does not include this value, so it will not sync or restore. Please wire it into cloud sync for Web/device parity.

The idea is good; once the TV navigation and real channel-number sorting are fixed, it will be much closer to merge-ready.

PR #518, opened 2026-08-06
fix: “Continue watching” does not display titles with the updated language

Aaronnn17:
I'm reopening the pull request because I was creating the previous one from my “main” branch and didn't realize it. I'm reopening this one from another appropriate branch.

I think I've fixed the issues you mentioned earlier, and I've tested it in Android Studio to make sure it doesn't crash during compilation.

Please forgive me for any other errors—I'm a total beginner at programming, and these are the first mini-projects I've ever done!


ProdigyV21:
Thanks for reopening this and addressing the earlier feedback. The PR now compiles, covers both Trakt and fallback items, and limits processing to 50 rows, so it is much closer.

I found two remaining issues before merge:

1. **Launcher refresh is still too expensive.** The final `selectedItems.map` calls full `getMovieDetails()` / `getTvDetails()` for every item and `getSeasonEpisodes()` for every TV episode, sequentially. These methods also do unrelated IMDb-rating and watched-state work. Because launcher refresh is called from several playback/history actions, this can generate many requests and make launcher updates slow. Please use lightweight TMDB calls that fetch only the localized title/episode title, with bounded concurrency and a small language-aware cache.

2. **A language change may still keep the old title.** `MediaRepository` caches details using only the media ID, not the language, and changing `contentLanguage` does not invalidate those caches. `setContentLanguage()` also does not refresh the launcher rows. Please make these title caches language-aware (or clear them when language changes) and call `launcherContinueWatchingRepository.refreshForCurrentProfile()` after changing the language.

Minor cleanup: `git diff --check` reports trailing whitespace on lines 103 and 173.

The exact merge into current main is clean, GitHub Build Check passes, and both Play/Sideload compiles and unit-test suites pass. The idea is useful; fixing the two points above should make it reliable and safe to merge.

Aaronnn17:
Hello again! After a few changes and tests, I think it's working as it should now. I've attached some screenshots showing the language switching between Spanish, Italian, French, and English—all in a matter of seconds.

image
image
image
image

(_Obviously, it only translates the titles that have been translated!_)

PR #523, opened 2026-08-07
fix(di): add @JvmStatic to AppModule @Provides methods for sideload flavor build

Himanth-reddy:
### Summary
Adds  to  methods inside  object.

### Problem
When compiling variant-specific builds like , Dagger Hilt fails with missing binding errors () because Kotlin  methods without  generate instance methods on  rather than static providers.

### Solution
Adding  to all  functions in  ensures static provider methods are emitted into Java bytecode, allowing Hilt Dagger annotation processing to resolve all bindings across all build variants.


Himanth-reddy:
Replaced by same-repository stacked PR #525

PR #563, opened 2026-08-14
fix(ui): restore instant predictive back commit and fix subpage exit latch

Himanth-reddy:
## Summary

This PR restores the 120fps fluid responsiveness to predictive back gestures and fixes the "duplicate/stale main page sliding out" visual glitch during sub-page dismissals.

---

#### 1. Input Lag & Double Animation from Blocking Settle Tween
* **Previous Behavior:** In `rememberArvioPredictiveBack`, `anim.animateTo(1f, tween(commitDurationMs))` was suspended *before* invoking `commit()`.
* **The Problem:** 
  1. This introduced a mandatory ~280ms blocking delay before navigation or dismissal actions could execute.
  2. Because Jetpack Compose containers (like `AnimatedVisibility`) have their own exit animations (`slideOutHorizontally`), delaying `commit()` caused the gesture to finish holding the screen in place, and *then* triggered a second, separate exit slide-out animation once `commit()` was eventually called.

#### 2. Stale/Duplicate Main Page Morphing During Subpage Slide-Out
* **Previous Behavior:** In `MobileSettingsLayout`, `AnimatedVisibility` observed `visible = page != "MAIN"`.
* **The Problem:** When `onNavigate("MAIN")` was triggered, `page` immediately became `"MAIN"`. During the 220ms exit slide-out animation, `MobileSettingsSubPage` and `mobileCategoryTitle` were recomposed with `page = "MAIN"`. This caused the exiting subpage to instantly morph into an empty / duplicate "MAIN" header on frame 0 and slide across the real main page sitting underneath it.

---

### Solution

1. **Instant Commit Execution in [ArvioPredictiveBack.kt](file:///mnt/combo/github/ARVIO/app/src/main/kotlin/com/arflix/tv/ui/motion/ArvioPredictiveBack.kt):**
   * Invoke `commit()` immediately upon gesture release so Compose's exit transitions take over fluidly with zero touch latency.
   * Concurrently complete the animation and synchronously reset state.

2. **Sub-Page Content Latching in [SettingsScreen.kt](file:///mnt/combo/github/ARVIO/app/src/main/kotlin/com/arflix/tv/ui/screens/settings/SettingsScreen.kt):**
   * Added `displayedSubPage` state tracking to preserve the active subpage's title and rows throughout the entire 220ms exit slide-out animation. The sub-page now flies off cleanly to the right while revealing the main settings page underneath.

---

### Why We Should Keep This Implementation

* **Zero Input Latency:** Eliminates the ~280ms delay between lifting your finger and route dismissal.
* **No Visual Collision:** Prevents double-animations where a screen pauses and then slides out a second time.
* **Clean State Transitions:** Resolves the flashing duplicate main page glitch during back navigation.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/motion/ArvioPredictiveBack.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/motion/ArvioPredictiveBack.kt
@@ -47,8 +47,8 @@ class ArvioBackMotion internal constructor() {
 /**
  * PredictiveBackHandler with real settle animations.
  *
- * Commit: animates the remaining travel to 1f *before* invoking [onCommit], so releasing
- * at 80% is never a jump cut. Cancel: animates back to 0f.
+ * Commit: invokes [onCommit] immediately while the visual state settles to 1f in the
+ * composition scope. Cancel: animates back to 0f.
  *
  * The settle is launched on [rememberCoroutineScope] because on cancel the handler's own
  * coroutine is already cancelled — suspending in the catch block would throw immediately.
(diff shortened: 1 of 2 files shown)


Himanth-reddy:
https://github.com/user-attachments/assets/8ae30550-b575-4768-83e0-c9b3a027ddae

PR #583, opened 2026-08-17
fix(player): finalize playback before next-episode navigation

### LINKED ISSUE #582: [BUG]: Episodes played using "Next Episode" may remain marked as unwatched
### Describe the Bug
Episodes watched while binge-watching a TV series can remain marked as unwatched when playback is continued using the **Next Episode** button from the player.

The affected episodes still appear as unwatched in the series details screen and, when spoiler blur is enabled, their thumbnail/synopsis remains hidden as a spoiler even though the episode has already been watched.

This seems specific to episodes transitioned through the player's **Next Episode** action rather than episodes started directly from the Details screen.

There may be a race condition in the Player → Next Episode navigation path: `onPlayNext()` immediately navigates to a new Player destination and removes the current Player from the back stack, while the final `saveProgress()` / watched-state update runs asynchronously in the current `PlayerViewModel`'s `viewModelScope`.

### Steps to Reproduce
1. Open ARVIO on Android TV / Google TV.
2. Open a TV series.
3. Start an episode from the series Details screen.
4. Watch the episode and use the player's **Next Episode** button to continue to the following episode.
5. Continue watching one or more episodes this way.
6. Return to the series Details screen.
7. Notice that some episodes watched through the **Next Episode** flow are still marked as unwatched.
8. With spoiler blur enabled, those episodes are also still displayed as spoilers.

### Expected Behavior
An episode that has reached the watched threshold or finished playback should be marked as watched before ARVIO navigates to the next episode.

Using **Next Episode** should produce the same watched/history state as finishing an episode and returning to the Details screen.

### Screenshots / Video
A screen recording can be provided if required.

---

#### **Platform (Check and complete the applicable section):**

- [x] **Android TV / Google TV / Fire TV**
  - **Device Model:** Nvidia Shield
  - **OS / Firmware:** [complete]
  - **App Version:** [complete]
  - **Install Source:** [complete]

- [ ] **Android Mobile / Tablet**

- [ ] **ARVIO Web (iOS / iPadOS / Desktop Browser)**

---

### Affected Component / Feature
- [x] **Playback / Player** (ExoPlayer/Media3, video stutter, audio passthrough, HDR/DV)
- [ ] **Subtitles / Audio** (Subtitle sync, missing tracks, AI subtitles)
- [ ] **IPTV / Live TV** (M3U / Xtream playlists, EPG guide, stream playback)
- [ ] **Home Servers & Sources** (Jellyfin, Emby, Plex, Telegram, Addons)
- [ ] **Catalogs & Metadata** (TMDB details, Trakt lists, MDBList)
- [x] **ARVIO Cloud / Profiles** (Cloud sync, watch progress, profile PINs, Trakt scrobbling)
- [x] **UI / Remote Navigation** (D-Pad focus issues, gestures, predictive back)
- [ ] **Crash / App Freeze**

---

### Media & Playback Details *(If reporting a video/audio/playback issue)*
- **Source Type:** Addon/Debrid
- **Stream Format / Codecs:** Not relevant / occurs independently of stream format
- **Player Settings:** Auto Play Next enabled

### Crash Logs / Logcat *(If applicable)*
No crash occurs.

---

### Additional Technical Context

The Player's manual Next Episode action calls `onPlayNext()` directly without explicitly finalizing the current playback session.

`AppNavigation` then navigates to the next Player route using:

```kotlin
popUpTo(Screen.Player.route) { inclusive = true }
```

Meanwhile, the final playback persistence performed during `PlayerScreen` disposal calls `PlayerViewModel.saveProgress()`, which starts a new coroutine in `viewModelScope`.

Since removing the current Player destination also destroys its ViewModel, that coroutine may be cancelled before `scrobbleStop()` and the watched-state update complete.

It may be safer to explicitly finalize/save the current episode before executing `onPlayNext()`, or otherwise make the final playback-state persistence survive destruction of the current Player ViewModel.

The manual **Next Episode** button, `MediaNext`, and the automatic post-episode countdown should ideally all use the same transition/finalization path to avoid different watched-state behavior depending on how the next episode is started.

Aerya:
Fixes #582

## Summary

Finalize the current episode's playback state before navigating to the next episode.

The existing Next Episode flow can replace the current Player destination before its asynchronous `saveProgress()` work finishes, which may cancel the watched/scrobble update with the old `PlayerViewModel`.

## Changes

- add `saveProgressAndWait()` to wait for the final playback save;
- route Next Episode transitions through a single helper;
- wait for persistence before calling `onPlayNext()`;
- cover the Up Next prompt, `MediaNext`, and the explicit Next Episode button;
- preserve the existing watched threshold behavior.


ProdigyV21:
Thank you for working on this. It addresses a real issue, and routing all Next Episode actions through one finalization path is the right direction.

I cannot merge it yet because a few lifecycle races remain:

- `saveProgressAndWait()` first cancels the active save. If that save already set `hasMarkedWatched = true` but is cancelled before the Trakt/history work finishes, the replacement save skips the watched block and the episode can still remain unwatched.
- The completion callback from an older cancelled job always clears `progressSaveJob`. It can therefore clear the reference to the newer save before `saveProgressAndWait()` joins it, allowing navigation to happen early.
- After the explicit save completes, Player disposal calls `saveProgress()` again. Because ExoPlayer may still report `isPlaying=true`, this can send a new scrobble-start for the old episode after it was already paused or stopped.

Please make `saveProgress()` return the exact new Job and join that Job directly, only clear `progressSaveJob` when it still references the completing Job, and prevent the disposal save from restarting/saving the old session during a successful Next Episode transition. A focused test covering an in-flight threshold save followed by Next Episode would also be valuable.

The PR otherwise merges cleanly into current main. GitHub CI passes, both Play and Sideload compile locally, all Sideload unit tests pass, and `git diff --check` passes. This is a useful fix once these final race conditions are handled.

PR #586, opened 2026-08-18
perf(build): reduce APK size via WebP assets, resource shrinking, and ARM-universal APK

Himanth-reddy:
## Summary of Changes

This PR delivers APK size optimizations and streamlines distribution into a single **ARM-Universal APK**:

### 📦 Build & Packaging:
• **Single ARM-Universal APK**: Restricted native ABIs to `armeabi-v7a` and `arm64-v8a` (32-bit and 64-bit ARM), removing ~60 MB of unused x86/x86_64 binaries while supporting 100% of physical Android TV and Fire TV devices with a single download.
• **Standard Gradle Packaging**: Removed ABI splits and custom repackaging tasks, letting Gradle build and sign `app-sideload-*.apk` natively without any variant cross-signing issues.
• **Font Integrity**: Kept the full Inter variable font to preserve all characters used by ARVIO's translations (Greek, Cyrillic, Vietnamese).

---

### 🚀 Optimizations Included:
• **Resource Shrinking**: Enabled `isShrinkResources = true` for release/staging builds.
• **ProGuard / R8 Log Stripping**: Enabled `-assumenosideeffects` rules in `app/proguard-rules.pro` for release builds.
• **WebP Asset Conversion**: Converted 111 drawables (sports cards, rank banners, avatars) from PNG/JPG to WebP (saving ~3.3 MB / 47.1% reduction).

---

## Verification & Measurements
• `./gradlew testSideloadDebugUnitTest` passes 100%.
• `./gradlew assembleSideloadDebug` builds cleanly and produces `app-sideload-debug.apk` containing only `arm64-v8a` and `armeabi-v7a`.
• Release APK size is reduced from ~184 MB down to **~65 MB**.


ProdigyV21:
Thanks for working on this — the WebP conversion and resource shrinking are useful, and the size reduction is significant. However, I think distributing one universal APK for all devices would be much easier for users than asking them to choose between several CPU versions.

Before merging, please fix these points:

1. The custom ARM packaging task scans all existing APK outputs. I reproduced a release build re-signing the existing debug ARM APK with the release key; the reverse could also leave a release APK debug-signed and cause update/install failures. Please scope packaging to the current variant only, preferably using Gradle's normal APK generation instead of manually rewriting/re-signing ZIP contents.
2. The subsetted Inter font removes 137 characters currently used by ARVIO's Greek, Cyrillic and Vietnamese translations. Please include all characters used by the shipped languages, or keep the full font.
3. Update the README because `app-sideload-debug.apk` is no longer generated.

The builds and unit tests pass, and the image conversion looks good. Once these points are fixed, this should be a worthwhile improvement, with one universal download being the clearest option for users.

Himanth-reddy:
I agree with the points about the packaging task, font subsetting, and README — those should definitely be fixed before merging.

For the universal APK though, I think we can meet the usability concern without shipping the x86/x86_64 binaries. The PR already generates an `arm-universal` APK containing both `armeabi-v7a` and `arm64-v8a`, and it's around 65 MB release compared to ~125 MB for the full universal APK.

Since ARVIO is primarily targeting Android/Android TV devices, ARM is by far the relevant architecture for actual users. x86/x86_64 are mainly useful for emulators and development environments. Shipping those binaries in the default universal download means every ARM user carries ~60 MB of architecture-specific native code they'll never execute.

So I think the best approach would be to make the ARM-universal APK the single recommended/default download. Users still get the "download one APK and don't worry about the CPU architecture" experience you're looking for, while we avoid the x86 dead weight.

We can still keep the individual `arm64-v8a` / `armeabi-v7a` APKs for people who specifically want the smallest possible package, and x86/x86_64 builds can remain available for emulator/developer use if needed.

That gives us:

* One simple APK for normal users
* ARM 32-bit + 64-bit compatibility
* ~65 MB release instead of ~125 MB
* No unnecessary x86/x86_64 native libraries for real devices
* Optional ABI-specific builds for users who want them

So I'd suggest keeping the general multi-ABI approach from the PR, but making `arm-universal` the primary distribution rather than the full 4-ABI universal APK.

Himanth-reddy:
The poll has ended, and the users chose to go with a universal APK. I’ll update the PR with a new commit accordingly.

Himanth-reddy:
Following the Discord community poll and the review feedback, this PR has been updated:

1. **Single ARM-Universal APK**: Restricted the target ABIs to `armeabi-v7a` and `arm64-v8a` (no x86/x86_64 binaries). Users get a single APK download that works across all physical Android TV and Fire TV devices with zero architecture selection required, while reducing the release size to ~65 MB (down from ~184 MB).
2. **Clean Gradle Generation**: Removed ABI splits and custom repackaging tasks. APKs are generated and signed directly by Gradle as standard `app-sideload-*.apk` artifacts, avoiding any variant cross-signing concerns.
3. **Full Font Preserved**: Kept the full Inter font to ensure Greek, Cyrillic, and Vietnamese characters remain intact.
4. **README & CI Consistency**: Retains the standard `app-sideload-debug.apk` paths in `README.md` and CI workflows.

PR #591, opened 2026-08-21
feat(settings): Add Stalker portal configuration for live TV

ReichiMD:
## Summary
Adds a settings UI to configure a Stalker/Ministra IPTV portal with a live-TV playlist, plus a client that resolves the correct API base path and handles pagination edge cases. Channels are integrated into the existing IPTV system.

### Key Changes
- **StalkerApi**: probes portal root, /stalker_portal, /portal, /ministra for a valid API base; pagination stops when portals ignore page parameters
- **SettingsScreen**: "Add Stalker portal" row in the IPTV section (mobile + TV); input modal for URL + MAC
- **Strings**: German and English labels/placeholders/help text
- **Tests**: new StalkerApiTest with 3 tests covering handshake probing and token failure

### Verification
- Local: ./gradlew testSideloadDebugUnitTest -> BUILD SUCCESSFUL (all 38 tasks)
- Tested with a01.live portal: 2116 channels, 44 groups, channel switching ~2s

---

This PR was created by an AI agent (OpenHands) on behalf of @ReichiMD.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt
@@ -11,7 +11,7 @@ import java.util.concurrent.TimeUnit
  * Stalker/Ministra portal API client for MAC-based IPTV authentication.
  * Converts Stalker portal channels into the same IptvChannel format as Xtream/M3U.
  */
-class StalkerApi(
+open class StalkerApi(
     private val portalUrl: String,
     private val macAddress: String
 ) {
@@ -160,7 +160,7 @@ class StalkerApi(
         }
     }
 
-    private fun doGet(url: String): String {
+    open fun doGet(url: String): String {
         val builder = Request.Builder().url(url)
         baseHeaders.forEach { (k, v) -> builder.addHeader(k, v) }
         val response = client.newCall(builder.build()).execute()
--- a/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt
@@ -105,6 +105,7 @@ open class StalkerApi(
     /** Step 3: Get all channels */
     suspend fun getChannels(): List<IptvChannel> {
         val channels = mutableListOf<IptvChannel>()
+        val seenChannelIds = HashSet<String>()
         try {
             // Get genres first for group names
             val genreUrl = "$apiBase/server/load.php?type=itv&action=get_genres&JsHttpRequest=1-xml"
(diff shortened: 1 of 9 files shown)


ProdigyV21:
Great pr, please take a look at this, then I can merge!

Findings
1. [P1] Stalker is ignored when a normal IPTV playlist exists
   [IptvRepository.kt:1591](https://github.com/ProdigyV21/ARVIO/blob/578efa1477b5bc16cb5334fbf1d5b61564f10c4a/app/src/main/kotlin/com/arflix/tv/data/repository/IptvRepository.kt#L1591) only loads Stalker when m3uUrl is blank. Users can configure it alongside playlists, but its channels silently never appear. Stalker channels must be merged with active playlist channels in both normal loading and startup prefetch.
2. [P1] Common /c/ portal URLs can select the wrong API path
   [StalkerApi.kt:74](https://github.com/ProdigyV21/ARVIO/blob/578efa1477b5bc16cb5334fbf1d5b61564f10c4a/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt#L74) considers almost any non-<html> response valid. A typical <!DOCTYPE html> 404 page or empty response can therefore stop probing early, causing the real endpoint to be missed. Only accept a candidate when the handshake response contains a valid token, and add tests for /c/ and /stalker_portal/c/.
3. [P2] A Stalker portal cannot be removed separately
   The dialog requires both fields, so users can edit it but cannot clear it without deleting all IPTV configuration. Add a dedicated remove action or allow both blank fields to clear only Stalker.
The feature itself is useful and well presented on mobile and TV. It imports live channels from Stalker/Ministra subscriptions using a portal URL and MAC address; VOD and EPG are explicitly not included yet.

ReichiMD:
Thanks for the thorough review! All three points make sense — I'm working on them now:

1. **Hybrid loading**: Stalker channels will be merged with active playlist channels — parallel loading with `stalker:`-prefixed IDs to avoid collisions, in both normal loading and startup prefetch.
2. **API path validation**: `resolveApiBase` will only accept candidates whose handshake response contains a valid token, plus new tests for `/c/` and `/stalker_portal/c/` portal URLs.
3. **Removable portal**: Adding a dedicated way to clear the Stalker configuration without touching playlist settings.

I'll update this PR once I've verified the changes on my device. 👍

ReichiMD:
## Review findings — all fixed and verified on device

Thanks for the detailed review. All three findings are addressed in the latest commit (`b5189214`) and verified on a real Android TV device with both an M3U playlist and a Stalker portal configured simultaneously (hybrid mode).

### Finding 1 — Stalker ignored when an M3U playlist is configured [P1]
Stalker channels are now loaded **in parallel** with M3U playlists and merged into the channel list. Both sources load and play correctly in hybrid mode.

### Finding 2 — `/c/` portal URLs hit the wrong API path [P1]
Portal path probing is now strict: it probes `/`, `/stalker_portal`, `/portal`, and `/c/` in order and **stops as soon as a response contains a valid JSON token**, ignoring HTML 404 pages or empty bodies. Unit tests cover each path variant (`StalkerApiTest`).

### Finding 3 — Stalker not separately removable [P2]
Clearing the portal URL + MAC fields now removes **only** the Stalker config. The "Delete TV Playlists" button removes all sources (M3U + Stalker).

### Additional fixes found during on-device testing
- **Playback in hybrid mode:** `cachedStalkerApi` was never set in hybrid mode, so `create_link` failed and Stalker streams would not play. Fixed.
- **Channel-count toast:** the add toast now shows the merged total (M3U + Stalker), and a duplicate toast that overwrote it was removed.
- **Delete-all button:** now also clears the Stalker portal + MAC so it truly removes all sources.
- **TV D-pad navigation:** horizontal Left/Right navigation on IPTV playlist rows was broken after the Stalker row was moved next to "Add playlist" — the D-pad guards were checking a stale index range. Fixed and verified on TV.

No debug logging, test workflows, or project-internal notes are included in this PR.

---

_This comment was created by an AI agent (OpenHands) on behalf of @ReichiMD._

PR #603, opened 2026-08-24
feat(iptv): group Live TV categories by playlist order

Saelon600:
## Summary
- Groups Live TV categories under their configured IPTV playlist/provider instead of flattening similarly named groups together.
- Preserves the saved playlist order in provider filters, TV category navigation, and touch category navigation.
- Keeps configured playlists visible when a large paged channel window currently contains channels from only one provider.
- Adds focused coverage for playlist ordering, category isolation, and paged-count behavior.
- Preserves unmatched Stalker and other-source categories in dedicated sections for hybrid configurations.
- Keeps the legacy flat category list when only one populated source exists.

## Why
Multi-playlist setups can contain the same category names from different providers. Flattening those categories makes navigation ambiguous and can mix channels from the wrong playlist.

## Testing
- `PlaylistCategorySectionsTest`
  - sections follow configured playlist order
  - categories do not mix between playlists
  - provider filters preserve saved reorder
  - paged counts retain all configured playlists
  - hybrid playlist + Stalker categories remain visible
  - single-source navigation remains flat without a collapsed parent
- `./gradlew :app:compileSideloadDebugKotlin`
- `./gradlew :app:testSideloadDebugUnitTest`
- `git diff --check`

## Scope
Android Live TV provider/category navigation only. No EPG actions, VOD lookup, playback rules, responsive layout, backend, or app version changes.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/tv/live/CategorySidebar.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/tv/live/CategorySidebar.kt
@@ -97,6 +97,7 @@ import kotlinx.coroutines.launch
 fun CategorySidebar(
     tree: LiveCategoryTree,
     selectedId: String,
+    playlistSections: List<PlaylistCategorySection> = emptyList(),
     expanded: Boolean,
     listState: LazyListState,
     focusRequester: FocusRequester? = null,
@@ -124,6 +125,7 @@ fun CategorySidebar(
     )
     var expandedCountry by rememberSaveable { mutableStateOf<String?>(null) }
     var expandedAll by rememberSaveable { mutableStateOf(false) }
+    var expandedPlaylistIds by rememberSaveable { mutableStateOf(emptyList<String>()) }
     var activeMenu by remember { mutableStateOf<CategoryMenuState?>(null) }
     val searchFocusRequester = remember { FocusRequester() }
     val selectedCategoryFocusRequester = remember { FocusRequester() }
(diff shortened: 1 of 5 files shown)


ProdigyV21:
Thanks for this contribution. Grouping categories by playlist is a useful improvement for multi-playlist setups, and the ordering/isolation logic looks good. The exact merge also compiles and the focused tests pass.

There are two things to fix before merging:

1. **Hybrid playlist + Stalker setups lose Stalker categories.** `buildPlaylistCategorySections()` only includes IDs from `config.playlists`, but once any playlist section exists, the TV and touch UIs replace the complete global category list with those sections. Stalker channels use the `stalker:` source ID and are not in `config.playlists`, so their groups disappear from category navigation. Please preserve unmatched categories in a Stalker/Other section, or model Stalker as its own section, and add a hybrid-source test.

2. **A single playlist now adds an unnecessary collapsed parent.** Users with only one playlist must open that playlist row before reaching their categories. Please either apply this grouping only when multiple sources are configured, or automatically expand the sole section.

Once those are handled, this should be a strong improvement and ready for another check.

Saelon600:
Addressed both review items in `cd254013`:

- Hybrid playlist + Stalker/other categories are now preserved as explicit unmatched-source sections, without changing the original category playlist IDs.
- A single populated source now returns to the legacy flat category list, so users do not get an unnecessary collapsed parent.
- Added focused hybrid Stalker and single-source regression tests; the full Sideload unit suite and Kotlin compilation pass.

PR #607, opened 2026-08-24
feat(iptv): support multiple configurable Stalker portals with per-portal groups

ReichiMD:
## Summary

Adds first-class support for configuring **multiple Stalker/Ministra portals** (up to 3) alongside existing M3U playlists, so each portal's category groups can be managed independently in the Settings UI — hide/show, reorder, bulk toggle — exactly like M3U playlist groups. Previously Stalker was limited to a single portal exposed only through a config dialog, with no group management.

### What changed

- **Multi-portal Stalker section (TV + mobile):** Stalker portals get their own settings section, separate from M3U playlists. Each portal is rendered as a row with a full action-chip set (categories / enable-disable / edit / rename / move up / move down / delete), mirroring the M3U row layout. An "Add Stalker portal" button is capped at 3 portals and disables when full.
- **Independent groups per portal:** Stalker channels use the `stalker:<portalId>:<origId>` id shape. Each portal's hidden groups and group order are stored under the portal's own id (e.g. `stalker1|group`) via the existing `PlaylistGroupKey` infrastructure, so managing one portal never affects another.
- **Rename dialog:** Portal name is editable (defaults to "Portal N"); subtitle is the portal URL.
- **Categories dialog per portal:** Reuses the M3U group screen, with a per-portal "Show all / Hide all" bulk toggle. Groups are only hidden/shown, never deleted — consistent with M3U behavior.
- **Enable/disable, sort, delete:** Toggling a portal drops its channels from the grid; up/down reorders the section; deleting a portal also clears its persisted group preferences (hidden groups + order) so re-adding starts clean.

### Bug fixes (pre-existing / discovered during development)

1. **TV grid ignored hidden groups and group order (all sources).** `hiddenGroups`/`groupOrder` are stored as `playlistId|groupName` keys, but `TvViewModel.buildPreparedGroups` and `currentVisiblePlaylistGroups` compared them against plain group names — so hiding/ordering had no effect on the TV view. Both now extract the group name from each key before comparing.
2. **Touch UI hid Stalker groups by the literal `stalker` prefix instead of the portal id.** The touch channel list derived the playlist id for hidden-group keys via `channel.id.substringBefore(':')`, which yields the literal `stalker` prefix for `stalker:<portalId>:<origId>` ids and therefore never matched the stored `stalker1|group` keys — hidden Stalker groups stayed visible on phones. A new helper resolves the portal id for Stalker ids (reusing `StalkerPortalSupport.portalIdFromChannelId`) and falls back to `substringBefore(':')` for M3U/Xtream, applied in `buildCategoryIndex`/`buildCategoryTree`/`buildPagedStartupChannelState` and the search filter. Category-id / sidebar keys are left unchanged.

### D-pad navigation

All hardcoded index dispatch maps in `SettingsScreen` were updated for the new Stalker rows and the categories bulk-toggle row. Helpers:
- `iptvRowMaxAction() = 5`
- `firstIptvGroupIndex(playlistId, groups, stalkerPortalIds)`
- `sectionMaxIndex("iptv") = 4 + playlists.size + stalkerPortals.size`

On-device TV testing confirmed correct focus, OK dispatch and left/right guards across all rows.

### Files changed

| File | Change |
|------|-------|
| `StalkerPortalSupport.kt` (new) | `STALKER_PLAYLIST_ID`, `MAX_STALKER_PORTALS`, `normalizeStalkerPortalEntry`, `portalIdFromChannelId`, encode/decode |
| `IptvRepository.kt` | `saveStalkerPortals` (normalize, cap at 3, clear prefs of removed portals); legacy `saveStalkerConfig`/`clearStalkerConfig` retained |
| `SettingsViewModel.kt` | Stalker CRUD: add/edit/toggle/move up/down/remove/rename/manage categories; `persistStalkerPortals()` private |
| `SettingsScreen.kt` | Stalker section (TV + mobile), rename dialog, per-portal categories + bulk toggle, D-pad index helpers & dispatch |
| `TvViewModel.kt` | Fix `PlaylistGroupKey` mismatch in hidden-group/order filtering |
| `LiveCategory.kt` / `LiveTvScreen.kt` | Fix hidden-group key to use portal id for Stalker on touch UI |
| `strings.xml` | Section titles, "Portal N" template, content descriptions, bulk-toggle labels |
| Tests | `StalkerPortalSupportTest` (normalize), `StalkerDpadIndexTest` (dpad indices) |

### Compatibility / scope notes

- No portal URLs, MAC addresses, or tokens are included — only neutral placeholders/labels.
- The cache-snapshot bugfix from the prior session (`getMemoryCachedSnapshot`/`getCachedSnapshotOrNull`) is intentionally untouched.
- Fork-only CI helpers (`test-apk.yml` and `.github/test-signing/`) used during development have been removed from this branch; the upstream `build-check.yml` remains the gate.

---

_This PR was created by an AI agent (OpenHands) on behalf of @ReichiMD._


ProdigyV21:
This is a valuable feature and the overall direction looks good. I verified the exact merge against current main: Play and Sideload compile, the Sideload unit tests pass, and GitHub checks are green. However, I found a few functional cases that should be fixed before merging:

1. **Playback cache must include the portal ID.** `TvViewModel` currently caches resolved Stalker streams using only the raw command. Two portals can return the same command, causing one portal to reuse another portal's resolved stream.

2. **Portal IDs can be duplicated.** IDs are generated from the current list size. If Portal 2 is deleted from three portals, adding another creates a second `stalker3`. Please generate the first unused ID or use a stable UUID.

3. **Live TV categories are not fully isolated per portal.** Category IDs still use the shared `stalker` prefix, so matching group names from different portals are combined. Live-sidebar hide/reorder actions can also save `stalker|Group` while filtering expects `stalker1|Group`. The portal ID needs to be used consistently for category creation, matching and actions.

4. **Please complete legacy migration.** Existing channel IDs change from `stalker:<channel>` to `stalker:stalker1:<channel>`, but favourites, recents, last channel, hidden groups and ordering are not converted. Older cloud snapshots also use the removed single-portal fields and need backward-compatible import support.

5. **Cloud restore currently drops Stalker group ordering.** The valid source-ID set only contains M3U playlist IDs; include the normalized Stalker portal IDs as well.

Please add regression tests for two portals returning the same command, delete-middle-then-add, duplicate group names across portals, legacy upgrade, and cloud round-trip. Once these are covered, this should be a strong addition.

PR #609, opened 2026-08-25
fix: EPG hidden categories dropdown (v1.10.18, versionCode 326, versionName 1.10.17)

Saelon600:
Fixes EPG hidden categories dropdown not showing when there are hidden groups.

- Ensures dropdown appears even when all groups in a category are hidden
- Maintains existing behavior for visible groups

Related to versionCode 326 / versionName 1.10.17


Himanth-reddy:
I checked the PR and the 2.8M+ lines are mostly coming from reference/official-playstore-jadx/, which looks like the full decompiled Play Store APK.

For this PR, the actual change is only in LiveTvEnhancements.kt, LiveTvScreen.kt, and the Gradle config, so we shouldn't be adding the entire decompiled APK to the repository.

Can you remove reference/official-playstore-jadx/ from the PR and keep only the files/code that are actually required for the fix?

Also, please check reference/extract_official_build_config.py and make sure we aren't committing any extracted API keys, client secrets, or other credentials from the official build.

Saelon600:
> I checked the PR and the 2.8M+ lines are mostly coming from reference/official-playstore-jadx/, which looks like the full decompiled Play Store APK.
> 
> For this PR, the actual change is only in LiveTvEnhancements.kt, LiveTvScreen.kt, and the Gradle config, so we shouldn't be adding the entire decompiled APK to the repository.
> 
> Can you remove reference/official-playstore-jadx/ from the PR and keep only the files/code that are actually required for the fix?
> 
> Also, please check reference/extract_official_build_config.py and make sure we aren't committing any extracted API keys, client secrets, or other credentials from the official build.

Thanks for spotting that. Not sure what happened. Closed this and cleaned where possible, resubmitted a fresh PR for ease.

PR #611, opened 2026-08-26
fix: EPG hidden categories dropdown shows even when all groups in a category are hidden

Saelon600:
Fixes EPG hidden categories dropdown not showing when there are hidden groups.

- Ensures dropdown appears even when all groups in a category are hidden
- Maintains existing behavior for visible groups

Related to versionCode 326 / versionName 1.10.17

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/tv/live/LiveTvEnhancements.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/tv/live/LiveTvEnhancements.kt
@@ -74,6 +74,7 @@ data class PlaylistCategorySection(
 fun buildPlaylistCategorySections(
     config: IptvConfig,
     categories: List<LiveCategory>,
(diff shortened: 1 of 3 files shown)


ProdigyV21:
Thanks for working on this. The hidden-category fix itself is useful, but this cannot be merged yet.\n\nThe PR currently fails both Play and Sideload builds. LiveTvScreen removes the onNavigateToDetails parameter, uses outdated EPG callback signatures, and no longer passes clockTickMillis to FullscreenGuideOverlay.\n\nIt also unintentionally removes roughly 380 lines from the recently merged EPG actions work in #604, including Stream Now/Search Sources, temporal programme handling, VOD lookup, and the programme action dialog. Those removals are unrelated to this fix.\n\nPlease restore LiveTvScreen.kt from the latest main, then reapply only the small hidden-group changes in uildPlaylistCategorySections() and its call site. Please also add a test for the case where every group in a category is hidden. Once those are done, I can review it again.

PR #623, opened 2026-08-31
feat(stalker): add EPG guide support

ReichiMD:
Summary
Stalker/Ministra portal channels currently show no EPG guide data in Live TV (the guide grid stays empty), even though the portal API supports lightweight now/next endpoints. This adds a Stalker EPG fetch path and merges it for Stalker channels only, leaving M3U/Xtream EPG untouched.
Approach
Live-tested against two real portals (raw findings documented privately, not part of this PR): portal EPG behavior varies — some answer get_simple_data_table (ch_id=all), some only get_epg_info, and some only respond per-channel via get_short_epg. The implementation therefore tries three things, in order, per active portal:
StalkerApi.getEpg() — get_simple_data_table (bulk, one request per portal), falling back automatically to get_epg_info when the first comes back empty. Both known response shapes are handled (flat array, and the {"data": {"<ch_id>": [...]}} wrapper confirmed on a real ~25 MB portal response).
StalkerApi.getShortEpg(chId) — per-channel fallback for portals where both bulk endpoints come back empty, used for batches up to 100 channels (limited concurrency, matching the existing Xtream short-epg pattern). Oversized batches (a full catalog backfill can be thousands of channels) skip this and rely on the on-demand visible-guide refresh path instead, which already requests small batches.
The raw per-portal bulk result is cached for 5 minutes — the on-demand visible-guide refresh calls the fetch for every small visible batch (often every few seconds while scrolling/switching channels), and a large portal's response can take 1.5–5s to fetch; without caching this caused noticeable lag, confirmed live.
Scope
StalkerApi: getEpg() + getShortEpg() (+ response models), robust to multiple field-name variants and both known container shapes.
IptvRepository: fetch Stalker EPG per active portal and merge into the EPG now/next map for Stalker channels only — no change to M3U/Xtream EPG handling. Wired into both places that populate the guide: the automatic background backfill (startCompleteEpgBackfill, previously gated on M3U/Xtream-only fields and never fired for a Stalker-only setup) and the on-demand visible-guide refresh (refreshEpgForChannels, previously Xtream-only and silently dropped every Stalker channel).
Unit tests for URL building, response parsing (all confirmed real-world shapes), and the now/next merge/isolation behavior.
Testing
Unit tests: ./gradlew testSideloadDebugUnitTest (existing suite + new cases) green.
On-device (phone, TV mode): Stalker portal → Live TV guide now shows program info for Stalker channels, smoothly (no lag) after the caching fix; M3U/Xtream guide unaffected. Verified across two independent portals with different response behavior; a portal with genuinely no EPG data shows no guide entries for it (expected, no crash, other sources unaffected).
Generated by Claude Code

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/api/StalkerApi.kt
@@ -160,21 +160,69 @@ open class StalkerApi(
     }
 
     /**
-     * Fetch the portal's lightweight now/next EPG table for all channels in one call.
+     * Fetch the portal's now/next EPG data for all channels.
      * [date] uses `YYYY-MM-DD`; blank means "today" on the server side.
+     *
+     * Tries `type=epg&action=get_simple_data_table` first (lightweight, one flat
+     * list). Some Stalker/Ministra portal builds don't implement the `epg` type
+     * handler at all (confirmed on-device: 0-byte response, while `itv`/`stb`
+     * actions on the same portal work fine) - for those, falls back to
+     * `type=itv&action=get_epg_info`, which some portals return as a flat list
+     * like the first action and others as an object keyed by channel id.
      */
     suspend fun getEpg(date: String = ""): List<StalkerEpgProgram> {
+        val dateParam = if (date.isBlank()) "" else "&date=${java.net.URLEncoder.encode(date, "UTF-8")}"
+        val simpleTable = fetchSimpleDataTableEpg(dateParam)
+        if (simpleTable.isNotEmpty()) return simpleTable
+        return fetchEpgInfoFallback(dateParam)
+    }
+
+    private fun fetchSimpleDataTableEpg(dateParam: String): List<StalkerEpgProgram> {
         return try {
-            val dateParam = if (date.isBlank()) "" else "&date=${java.net.URLEncoder.encode(date, "UTF-8")}"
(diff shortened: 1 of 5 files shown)


ProdigyV21:
This is a great addition and the overall Stalker EPG approach looks solid. I tested the exact merge against current `main`: it merges cleanly, both Play and Sideload compile, `git diff --check` passes, and the Sideload unit suite is green.

There are three runtime issues I think should be fixed before merging:

1. **Large-response memory usage:** `get_epg_info` can return about 25 MB, but the code currently holds the response string, builds a full Gson tree, and then creates another model graph. On lower-memory TVs this can cause an OOM. Please use streaming/bounded parsing and cache compact per-channel guide data instead of every raw program.
2. **Cache isolation:** `stalkerEpgCache` is keyed only by IDs such as `stalker1` and is not cleared by `invalidateCache()`. After changing portals or profiles, data from the old portal can be reused for up to five minutes. Please bind the entry to the portal/API identity and clear it during invalidation.
3. **Short-EPG request repetition:** `get_short_epg` results are not cached. Portals without bulk EPG support can therefore receive dozens of requests whenever unresolved channels refresh. Please add a short per-portal/per-channel TTL, including for empty results.

The feature itself is valuable and close to ready; these changes would make it safe for lower-end TVs and larger installations.

ReichiMD:
All right, and thanks for fixing it right away.

PR #626, opened 2026-09-01
feat: Iptv toggles

Aaronnn17:
I've added the option to individually select the Live TV, Series, and Movies options in the IPTV lists. This is useful when you only want a certain type of content or, in my case, to have two lists: one exclusively for Live TV (from a list editor) and the other for series and movies directly from the provider with their corresponding TMDB IDs.

image

**Already tested with Android Studio**


ProdigyV21:
Thanks for adding this - per-playlist IPTV content toggles are a useful feature, especially for separating a Live-only list from an Xtream Movies/Series list. The exact merge builds successfully for both Play and Sideload, and the existing IPTV tests pass, but I found a few things that need fixing before merge:

1. Movie/series lookup still resolves only the primary/first Xtream playlist. If list 1 is an edited M3U Live-only list and list 2 is the Xtream provider with Movies/Series enabled, VOD lookup can return nothing. Please iterate the enabled playlists/providers and use the ones where `importVod` or `importSeries` is enabled.
2. `importLiveTv` is only checked inside the Xtream path. A normal M3U playlist still loads all channels when Live TV is disabled. Please apply the toggle before both Xtream and M3U loading.
3. The three new flags are not included in the IPTV cache signatures. A cloud-restored toggle change can therefore reuse stale persisted channels. Please include them in the relevant signatures or clear persisted channel caches whenever they change.
4. Please also remove the trailing whitespace in `IptvRepository.kt` and preferably move the three new labels into string resources.

A focused test with list A = Live-only and list B = Movies/Series-only would be useful to verify the main use case end to end. Once the first three items are fixed, this should be a good addition.

PR #627, opened 2026-09-01
fix encoding issue in subtitles

silentbil:
## fix(subtitles): detect subtitle file encoding instead of assuming UTF-8                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  ### Problem                                                                                                                                                                                                                                                                                                       
  Subtitles from providers that still serve legacy code pages rendered as rows of `?`                                                                                                                                                                                                                               
  (U+FFFD). `loadRaw` decoded every downloaded subtitle as UTF-8, and the damage was                                                                                                                                                                                                                                
  persisted into the local cache copy. Separately, tracks beyond the preload cap were                                                                                                                                                                                                                               
  handed to ExoPlayer as remote URLs, which media3 also decodes as UTF-8 — so a track                                                                                                                                                                                                                               
  could look fine on auto-select and break when re-picked from the menu.                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                    
  ### Fix                                                                                                                                                                                                                                                                                                           
  - `loadRaw` detects the encoding: BOM → strict UTF-8 attempt → legacy code page chosen                                                                                                                                                                                                                            
    by the subtitle's declared language (windows-1252 fallback). Strict decoding is the                                                                                                                                                                                                                             
    key: UTF-8 is self-validating, so legacy files fail cleanly instead of silently                                                                                                                                                                                                                                 
    producing U+FFFD.                                                                                                                                                                                                                                                                                               
  - `loadRaw`/`loadCues` take `lang`; passed at all four call sites.                                                                                                                                                                                                                                                
  - `selectSubtitle` serves a locally decoded copy instead of a remote URL — preloaded copy                                                                                                                                                                                                                         
    when available, else fetched (6 s bound, falling back to the remote URL as before).                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                    
  ### Tests                                                                                                                                                                                                                                                                                                         
  `SubtitleEncodingTest` (9): legacy decode, language-tag variants, both BOMs, UTF-8                                                                                                                                                                                                                                
  preferred over a legacy page that would also decode, unknown-language safety, cue timings                                                                                                                                                                                                                         
  preserved. Full suite green. Verified on device: every rendered cue reports `fffd=0`.

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerViewModel.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/screens/player/PlayerViewModel.kt
@@ -76,6 +76,13 @@ private const val PLAYBACK_DIAGNOSTICS = true
 // side-loaded at startup (each becomes a MediaItem SubtitleConfiguration read at prepare).
 private const val MAX_PRELOAD_SUBS = 15
 
+/**
+ * How long a manual subtitle pick waits for its local (correctly decoded) copy before falling back
+ * to the remote URL. Generous enough for a normal ~100 KB fetch, short enough that a stalled addon
+ * doesn't make the menu feel broken.
+ */
+private const val SUBTITLE_LOCALIZE_TIMEOUT_MS = 6_000L
+
 private fun playbackDiag(message: String) {
     if (PLAYBACK_DIAGNOSTICS) {
         System.err.println("[PlaybackDiag] $message")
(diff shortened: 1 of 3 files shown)


ProdigyV21:
Thanks for this contribution — the encoding detection is a useful fix, and the new tests cover the main legacy-encoding cases well. Before merging, could you please address two lifecycle issues?

1. `subtitleLocalizeJob` is only cancelled when another subtitle is selected. Please also cancel it when `loadMedia()` starts, when `disableSubtitles()` is called, and when another mode intentionally clears/replaces the selected subtitle. Otherwise a delayed download can turn subtitles back on or apply a subtitle from the previous episode. A session/generation check before `applySelectedSubtitle()` would make this robust.

2. The 6-second `withTimeoutOrNull` does not reliably stop the synchronous `OkHttp.execute()` call. A stalled request can continue until OkHttp's longer network timeout. Please make the request cancellation-aware (for example with async `enqueue()` plus `Call.cancel()`) or apply a real per-call timeout so manual selection falls back after the advertised 6 seconds.

The decoder itself looks good and the subtitle encoding tests pass. Once these two cases are handled, this should be ready.

silentbil:
> Thanks for this contribution — the encoding detection is a useful fix, and the new tests cover the main legacy-encoding cases well. Before merging, could you please address two lifecycle issues?
> 
> 1. `subtitleLocalizeJob` is only cancelled when another subtitle is selected. Please also cancel it when `loadMedia()` starts, when `disableSubtitles()` is called, and when another mode intentionally clears/replaces the selected subtitle. Otherwise a delayed download can turn subtitles back on or apply a subtitle from the previous episode. A session/generation check before `applySelectedSubtitle()` would make this robust.
> 2. The 6-second `withTimeoutOrNull` does not reliably stop the synchronous `OkHttp.execute()` call. A stalled request can continue until OkHttp's longer network timeout. Please make the request cancellation-aware (for example with async `enqueue()` plus `Call.cancel()`) or apply a real per-call timeout so manual selection falls back after the advertised 6 seconds.
> 
> The decoder itself looks good and the subtitle encoding tests pass. Once these two cases are handled, this should be ready.

 @ProdigyV21 Thanks — both were real, fixed.                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                    
  **1. Lifecycle.** `subtitleLocalizeJob` was only cancelled in `selectSubtitle`; it's now torn                                                                                                                                                                                                                     
  down in all six places `subtitleSelectionJob` is (`loadMedia`, `updatePlayerTextTracks`,                                                                                                                                                                                                                          
  `selectSubtitle`, `toggleLiveAudioTranslation`, `startMatchListening`, `disableSubtitles`) via a                                                                                                                                                                                                                  
  `cancelSubtitleLocalization()` helper. As you suggested, that helper also bumps a generation                                                                                                                                                                                                                      
  counter checked before `applySelectedSubtitle` — cancellation alone was racy, since the apply is                                                                                                                                                                                                                  
  a plain state write with no suspension point, so a job cancelled after its last suspension would                                                                                                                                                                                                                  
  still land. Stale results are now dropped.                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  **2. Timeout.** You're right that `withTimeoutOrNull` couldn't interrupt the blocking                                                                                                                                                                                                                             
  `execute()` — the client has no `callTimeout`, so a stall ran on to the 10 s connect / 15 s read                                                                                                                                                                                                                  
  timeouts. `loadRaw` now uses `suspendCancellableCoroutine` + `enqueue()` with                                                                                                                                                                                                                                     
  `invokeOnCancellation { call.cancel() }`, so the call is genuinely abortable — which also lets                                                                                                                                                                                                                    
  the preload pass cancel its 15 parallel downloads on teardown. Misleading KDoc corrected.                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                    
  That change surfaced a third issue: `runCatching` was swallowing `CancellationException`, which                                                                                                                                                                                                                   
  would have reported a cancelled download as a plain failure. `onFailure` now rethrows it.                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                    
  Two new tests use a `ServerSocket` that completes the handshake and never replies — the stall a                                                                                                                                                                                                                   
  slow addon produces. Both finish in ~0.7 s; with blocking `execute()` they'd take ~15 s and fail                                                                                                                                                                                                                  
  the `< 5 s` bound. No new test dependency needed.                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
  One gap to flag: the generation guard itself isn't unit-tested — that needs driving                                                                                                                                                                                                                               
  `PlayerViewModel` through Hilt across a `loadMedia`/`disableSubtitles` cycle, which the project                                                                                                                                                                                                                   
  has no harness for. Verified by inspection.

PR #633, opened 2026-09-03
feat(details): show age certification on movie and show details

ReichiMD:
### What

Movie and TV show details pages now show the age certification (FSK 16, PG-13, TV-MA, ...) next to the existing metadata line (rating - year - runtime), on both the phone and the TV layout.

### How

- The certification comes from TMDB through `append_to_response` on the detail calls that already run: `release_dates` for `/movie/{id}`, `content_ratings` for `/tv/{id}`. No new endpoint, no extra network request.
- The country is derived from `MediaRepository.contentLanguage` ("de-DE" -> "DE"), so no new setting is needed.
- Selection and labelling live in `util/ContentRating.kt`:
  - A movie can carry several certifications per country (theatrical, TV, streaming) that disagree. The theatrical one (`type` 3) wins, otherwise the lowest release type with a non-empty value.
  - An empty `certification` counts as absent - TMDB returns a lot of those.
  - Nothing for the content country -> the US value is used and marked as such (`US - R`). Nothing at all -> the badge is hidden entirely, no "unrated" placeholder.
  - Labelling: values that speak for themselves stay untouched (`R`, `PG-13`, `TV-MA`); bare numbers get their rating system (`16` in DE -> `FSK 16`) or else their country code (`12` in FR -> `FR 12`).

### Notes for review

- The rating-system names are deliberately kept in code and not in `strings.xml`: "FSK" is not a translation of "Rated", it is the name of the system a certification comes from. It follows the content country, not the interface language.
- Both new response fields are nullable with a default, so the shared movie/TV detail models stay valid when TMDB omits the block.
- `MediaItem.rating` is already used as a numeric score, so the new field is called `contentRating`.
- `append_to_response` is set as a default on the API interface, so it applies to every detail call. No behaviour change; the response is a few KB larger.
- Known, pre-existing limitation left alone on purpose: `detailsCache` is keyed without the content language, so after a language switch the certification can be up to 5 minutes stale - exactly as titles and overviews behave today.

### Deliberately not included

Kids-profile filtering / parental lock, MDBList as a second source, a settings toggle, and certifications on catalog cards. Each of those is its own topic.

### Testing

- 19 unit tests covering the selection and labelling rule (`ContentRatingTest`).
- Sideload debug build installed and checked on a phone (TV layout) and on a TCL C7K Android TV with the remote. The badge appears in both layouts and disappears cleanly when TMDB has no certification.

---

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/data/api/TmdbApi.kt
+++ b/app/src/main/kotlin/com/arflix/tv/data/api/TmdbApi.kt
@@ -64,6 +64,7 @@ interface TmdbApi {
     suspend fun getMovieDetails(
         @Path("movie_id") movieId: Int,
         @Query("api_key") apiKey: String,
+        @Query("append_to_response") appendToResponse: String = "release_dates",
         @Query("language") language: String? = null
     ): TmdbMovieDetails
 
@@ -71,6 +72,7 @@ interface TmdbApi {
     suspend fun getTvDetails(
         @Path("tv_id") tvId: Int,
         @Query("api_key") apiKey: String,
+        @Query("append_to_response") appendToResponse: String = "content_ratings",
         @Query("language") language: String? = null
     ): TmdbTvDetails
 
@@ -263,7 +265,10 @@ data class TmdbMovieDetails(
     val genres: List<TmdbGenre> = emptyList(),
     val status: String? = null,
     val adult: Boolean = false,
-    @SerializedName("belongs_to_collection") val belongsToCollection: TmdbCollectionRef? = null
+    @SerializedName("belongs_to_collection") val belongsToCollection: TmdbCollectionRef? = null,
+    // Appended via append_to_response. Nullable so the response stays valid
+    // when TMDB omits the block.
+    @SerializedName("release_dates") val releaseDates: TmdbReleaseDatesResponse? = null
 )
 
 /** Reference to a TMDB collection (franchise) returned inside movie/TV details. */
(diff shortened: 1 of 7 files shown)


ProdigyV21:
Got screenshot on how it looks?

ReichiMD:
Hi @ProdigyV21 - here is how it looks. Three screenshots of the same title, one per content language, so the labelling rule is visible in one go.

Screenshot_20260904-104011
Screenshot_20260904-103952
Screenshot_20260904-103936


The certification follows the *content* language, not the interface language, because it is data rather than translatable text. Values that speak for themselves stay untouched (`PG-13`, `R`, `TV-MA`), a bare number picks up its rating system (`16` in DE -> `FSK 16`), and formats like the French ones come through as TMDB delivers them. If the country has no certification the US value is shown and marked as such (`US - R`); if there is nothing at all, the badge is hidden completely.

One thing I would like to add afterwards, in a separate small PR: turning the plain text into a rounded pill like the existing "In Cinema" badge in the TV layout, optionally with a colour that matches the age. I left it out here on purpose to keep this PR small, and since it touches exactly the same two insertion points in `DetailsScreen.kt`, it only makes sense once this one is merged.

Two questions on that, whenever you have a minute:

- Colour, or just the rounded shape in a neutral grey? A colour is only trustworthy where the certification carries a real age number (DE, GB, BR, AU, NL, plus a small explicit table for the US letters). Everything else would stay neutral - I would rather be honestly colourless than wrongly colourful. Also: I would use our own palette, not a rebuild of the official FSK/BBFC colours, to stay clear of their branding.
- If colour: fixed values in code, or should they become tokens in `ArvioSkinTokens` so a future skin can move them?

If you have no preference, I will start with the neutral rounded pill - smallest change, no new colour decisions, and it can gain colour later.

---

PR #644, opened 2026-09-04
feat(details): put the age certification in a rounded chip

ReichiMD:
### What
The age certification on the movie and show details screens is now shown in a
small rounded chip instead of plain text in the metadata row, so it reads as a
certification rather than as one more number next to the runtime. The hero
preview on the TV home screen shows it too — it had the data available but was
not displaying it.

### How
- Reuses the existing "In Cinema" pill pattern from the details screen: Box +
  RoundedCornerShape(6.dp) + padding(horizontal 10.dp, vertical 4.dp), moved to
  `ui/components/ContentRatingChip.kt` so both screens share one definition.
- Neutral translucent background, no colour coding by age — deliberately, see
  below.
- Details screen: applied to both layouts (phone and TV). The separator in front
  of the certification is dropped; the chip separates itself.
- TV home screen hero: the certification is carried through `HeroDetailsSnapshot`,
  which already holds the runtime, IMDb rating and budget from the same details
  call — so this costs no additional request.

### Notes for review
- No new strings, no changes to the certification logic or the TMDB models —
  presentation only, apart from the one field added to the hero snapshot.
- On TV the details metadata row has a fixed width and the genre text ellipsises;
  the chip takes slightly more room than the plain text did. Checked on a TCL C7K.

### Deliberately not included
- No colour by age rating: the age is only unambiguous for a few numeric systems
  (FSK, BBFC, ClassInd, ACB, Kijkwijzer) and would have to be guessed for
  letter-based ones (R, PG-13, TV-MA). A wrong colour is worse than none.
- Not shown in the phone home carousel: its cards come straight from the category
  rows and carry no certification, so it would mean a details request per card.
  On the phone the certification is on the details screen.
- No conversion of the remaining metadata (genre, year, runtime) to chips.
- No skin/theme tokens.

### Testing
Built as a sideload debug APK and installed on a phone and on a TCL C7K
(Android TV). Existing unit tests unchanged and green.

PXL_20260904_205556080 MP
PXL_20260904_205541332 MP
Screenshot_20260904-224951

---

### DIFF
--- a/app/src/main/kotlin/com/arflix/tv/ui/components/ContentRatingChip.kt
+++ b/app/src/main/kotlin/com/arflix/tv/ui/components/ContentRatingChip.kt
@@ -1,0 +1,51 @@
+package com.arflix.tv.ui.components
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shadow
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.TextUnit
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.arflix.tv.ui.theme.ArflixTypography
+
+/**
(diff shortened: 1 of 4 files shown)


Himanth-reddy:
I like the idea of making the certification more distinct, but I don't think the current chip quite sits right with the existing UI. The metadata row currently has a pretty clean and consistent rhythm with the rating, year and runtime separated by dots, and the chip makes the certification feel a bit visually heavier than the rest. I think it might be worth exploring a treatment that gives the certification some distinction without introducing a full pill/chip into the metadata row.

ProdigyV21:
I agree with himmanth, it makes it a bit too heavy. Should be in the same rythm as the rest.

ReichiMD:
Yes, I agree with you and I'll take another look at it.👍

PR #668, opened 2026-09-07
feat(home): add customizable background trailer playback

### LINKED ISSUE #663: [FEATURE]: Add customizable background trailer playback on Home screen
### Is your feature request related to a problem? Please describe.

Yes. Currently, on the ARVIO home screen, when the user keeps the focus on a movie card for around 5 seconds, the trailer starts automatically and opens in a full-screen view.

It would be better to have an option to play the trailer in the background while keeping the user on the home screen instead of opening the full-screen player.

### Target Platform(s)

* [x] **Android TV / Google TV / Fire TV**
* [ ] **Android Mobile / Tablet**
* [ ] **ARVIO Web (iOS / iPadOS / Desktop Browser)**

### Feature Category

* [ ] **Player & Playback**
* [ ] **Subtitles & Audio**
* [ ] **IPTV / Live TV**
* [ ] **Integrations & Sources**
* [ ] **Catalogs & Metadata**
* [ ] **Profiles & Cloud Sync**
* [x] **UI / UX & Customization**

---

### Describe the solution you'd like

Add a customizable option to control how automatic trailers are played on the home screen.

When the user keeps the focus on a movie card for a few seconds, the trailer could play as a background video while keeping the user on the home screen, with the movie information and UI elements displayed over the video.

The setting could allow the user to choose between:

* **Full-screen** — current behavior.
* **Background playback** — play the trailer as a background video on the home screen.
* **Disabled** — do not automatically start trailers.

### Describe alternatives you've considered

One alternative would be to keep only the current behavior of opening the trailer in full-screen mode.

However, background playback would be preferable for users who want a browsing experience similar to other applications, such as Nuvio.

### Mockups / UI References *(If applicable)*

The desired behavior is similar to the background trailer playback used by Nuvio, where the video can play behind the interface elements without taking the user away from the home screen.

### Compliance & Policy Check

* [x] I confirm this request adheres to ARVIO's source policy.

### Additional Context

The goal is not to remove the current behavior, but to provide a customization option so each user can choose how they prefer automatic trailers to be displayed.

procopio1000:
## Summary

Implements customizable automatic trailer playback on the Home screen, addressing #663.

### Changes

* Adds background trailer playback on the Home screen without opening the full-screen player.
* Respects the existing `trailerAutoPlay` setting.
* Respects the existing `trailerDelaySeconds` configuration.
* Respects the existing `trailerInCards` mode.
* Preserves in-card trailer playback when that mode is enabled.
* Adds Hero/Backdrop trailer playback when `trailerInCards` is disabled.
* Stops the current trailer immediately when focus changes.
* Cancels stale asynchronous trailer operations to prevent delayed playback from previously focused items.
* Keeps D-pad navigation available while trailers are playing.
* Reuses a shared trailer player and releases resources when entering the main player.
* Preserves the existing Home catalog layout.

### Validation

* `compileSideloadDebugKotlin` ✅
* `testSideloadDebugUnitTest` ✅
* `assembleSideloadDebug` ✅
* Tested on a physical Android TV via ADB ✅
* Verified trailer activation, delay, focus changes and playback behavior ✅

Closes #663


ProdigyV21:
got screenshots/screenrecordings of how this looks?

procopio1000:
> got screenshots/screenrecordings of how this looks?

Sure! I’ll record some screenshots/screen recordings right now and add them to the PR.

procopio1000:
> got screenshots/screenrecordings of how this looks?


Uploading 20260907-114051_IzjgOqBX.mp4…

procopio1000:
> got screenshots/screenrecordings of how this looks?

IMG_20260907_124736_296
IMG_20260907_124736_286
IMG_20260907_124736_110

procopio1000:
> got screenshots/screenrecordings of how this looks?


https://github.com/user-attachments/assets/3257d4d9-037a-43f7-b470-cbca230fb818

ProdigyV21:
Thanks for the contribution and the recordings! Optional trailer previews could be a nice addition, but this needs a few changes before merging.

**Please keep Home's existing appearance unchanged.** I don't want the smaller top-right backdrop or the new hero layout. Keep the current backdrop size, positioning, image alignment, padding/spacing and gradients, including the logo/description placement. Trailers should fit the existing layout, and with autoplay off Home should look exactly as it does on main.

The remaining issues are:

1. **YouTube extraction is restored.** Current main deliberately removed direct YouTube extraction from both Android distributions. This PR brings back `YouTubeExtractor`, direct stream playback and separate audio/video handling, and the existing copyright-safeguard test fails. Please keep those safeguards and use an official YouTube playback approach that meets its requirements, or trailer files we have permission to play directly.
2. **The trailer controls are still hidden.** The visible TV settings row map excludes autoplay, sound, delay and in-card mode. Fresh installs default to autoplay off with no way to enable it, while existing users cannot change their saved preferences. The controls need to be wired up for the agreed playback approach.
3. **The shared player cannot recover after activity destruction.** `MainActivity.onDestroy()` permanently releases the application-wide `TrailerPlayerPool`. Exiting and reopening while the process remains alive then reuses a disabled pool. Make its lifecycle restartable or scope it appropriately.
4. **In-card artwork does not return when playback ends.** `FeaturedMediaCard` hides the artwork after the first frame but never restores it when the trailer finishes or fails. Handle `onPlayingChanged(false)` so the card returns to its normal artwork instead of remaining black.

After updating, please rerun the build/tests and check TV navigation, exit/reopen, and trailer completion/error behavior. I'd like to keep the preview idea, without redesigning Home or undoing the recent compliance changes.

PR #688, opened 2026-09-11
feat(player): faster subtitle auto-match with AI text verification

silentbil:
## Summary                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  Find Best Match used to hide subtitles for the whole scan (up to ~20 s) and trusted timing overlap                                                                                                                                                                                                                
  alone, which can't tell a mistimed right subtitle from a wrong one.                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                    
  ## Changes                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  - **AI users:** AI translation holds the screen during the scan. The sync reference is read from the                                                                                                                                                                                                              
    running player's buffer of the embedded English track AI translates from, so there's no hidden                                                                                                                                                                                                                  
    subtitle and no second player. Reading waits until the player has actually switched to that track.                                                                                                                                                                                                              
  - **AI text verification:** candidate lines are paired with the reference by meaning. Timing-only                                                                                                                                                                                                                 
    matches with no paired lines are rejected, and offsets are confirmed by the pairing.                                                                                                                                                                                                                            
  - Only the top candidate is parsed; the rest load only if it fails.                                                                                                                                                                                                                                               
  - Offsets apply live in the text renderer (no MediaItem rebuild). The menu shows `· fixed +1.0s`.                                                                                                                                                                                                                 
  - A match is remembered only after 2 min of playback with it.                                                                                                                                                                                                                                                     
  - One "Adjusting subtitles…" indicator; ranking handles URL-encoded stream names.                                                                                                                                                                                                                                 
  - **Without AI:** unchanged (in-player reference).                                                                                                                                                                                                                        
       
  ## Why no second player                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                    
  A headless reference player (the #625 approach) was built and removed. On the same episode it pushed                                                                                                                                                                                                              
  peak Java heap from 276 MB to 536 MB, because reaching a muxed subtitle track means demuxing the video.                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                    
  ## Testing                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  - Unit tests pass.                                                                                                                                                                                                                                                                                                
  - Emulator:                                                                                                                                                                                                                                                                                                       
    - From S01E10, S02E05, House of the Dragon                                                                                                                                                                                                                                                                      
    - The Office: AI rejected three wrong subtitles that scored 0.81–0.86 on timing                                                                                                                                                                                                                                 
  - Release builds of `main` and this branch compared on a Googge TV Streamer.


ProdigyV21:
Thanks for the improvement! One small fix before merging: failed or inconclusive AI checks currently still add the subtitle to aiVerified. Please only mark it verified when the AI actually confirms the dialogue. Timing-only fallback is fine, but keep it separate so an unconfirmed result isn't remembered as a verified match. Please also add a test for failed/inconclusive AI responses.

silentbil:
@ProdigyV21 
Thanks, good catch. Fixed: only an AI-confirmed dialogue now counts as verified and gets remembered. A failed or inconclusive AI check still lets the subtitle win on timing, but it's no longer cached, so the next                                                                                            
playback scans again. Added AiVerdictTest for the failed, inconclusive, confirmed and rejected cases.

PR #715, opened 2026-09-17
Live TV guide: a calmer layout on TV, and the same guide on phone and tablet

trailblazestudio:
The live-TV guide was dense. Every row carried a logo, a name, a number, a
quality badge, a chrome bar and a second line of programme text, and the rows
shrank to make it all fit. At three metres that reads as noise.

This takes things away rather than shrinking them. Programme cells carry only a
title; the quality sits as a raised suffix after the channel name; the date
moves into the ruler's left stub; the chrome bar is gone. Rows land at 38dp —
roughly where they already were — but each one holds far less. Entering a group
folds the group column away and gives the guide the full width, and Back or
Left brings it back.

Phone and tablet get the same guide in the shape each screen wants: the phone
drags a bottom sheet up over the player, the tablet keeps a fixed 220dp group
column beside the grid with no landing page in between.

## Commits

- **`fix(live): follow the accent colour chosen in Appearance`** — the guide was
  the one screen in the app with its own colour. `MainActivity` only read the
  `accent_color` key, so a profile carrying the older `focus_border_color`
  resolved to null and all 41 highlights in `live/` fell back to hardcoded cyan.
- **`feat(live): move the sports match guide next to the sports channels`** —
  the match guide sat at the top of the group list as "Sports" while the genre
  filter "Sports · Global" sat far below inside the accordion. They now sit
  together, match guide first. Only the channel list carries a count.
- **`feat(live): a calmer, roomier TV guide, …`** — the layout itself, the
  phone and tablet shapes, and the settings that drive them.

## Settings

Settings → Sources → TV Guide: **New Guide Design**, **Guide Rows** (6–10, TV
only), **Channels from All Providers**. The switch lives in the TV settings and
cannot be reached from a phone, so touch devices always get the new layout —
asking there would have produced a new frame around old cells.

## Notes for review

- `LiveGuideDensity` is the whole measurement spec as plain Kotlin with no
  Compose imports, so it is unit-testable instead of eyeballed on a device. It
  computes against a 540dp canvas, so "Guide Rows" lands as intended on
  1080p-class screens and is approximate elsewhere.
- Two pre-existing problems surfaced while wiring the touch layouts: the touch
  branch hardcoded `compact = true`, so a 1280dp tablet ran the phone's stacked
  layout and the video filled the content area; and `TouchCategoryRail` had been
  sitting finished and uncalled. Both are fixed here.
- Rows with children ("All Channels", the countries) now expand on click instead
  of entering the group. Previously they did both, and the guide covered the
  column before you could reach the children. This applies to the old design
  too, as it is a fix rather than a variant.

## Verified

- 1343 unit tests green; `compileSideloadDebugAndroidTestKotlin` green; all three
  commits compile individually.
- On a 960×540dp TV emulator, by `uiautomator` dump: both sports rows in the
  accordion in the right order, search → pick channel closes the group column,
  Back reopens it.
- Phone (411×914dp) and tablet (1280×800dp) emulators: sheet drags both ways,
  8 rows collapsed; tablet shows the group column with 9 rows at 51dp.

Two tests are flaky and untouched by this work — they failed and then passed on
re-run: `EpgRefreshRequestsTest.archiveWaitsForShortGuideWithoutStartingConcurrentRequests`
and `DiscoverGridTest.anAgeRatingIsNeverSentForSeriesBecauseTmdbHasNoSuchFilter`.


ProdigyV21:
Got screenshots?

trailblazestudio:
Lukket — sendt fra den forkerte konto. Kommer igen fra den rigtige.

PR #716, opened 2026-09-17
Live TV guide: a calmer layout on TV, and the same guide on phone and tablet

GAPP99:
The live-TV guide was dense. Every row carried a logo, a name, a number, a
quality badge, a chrome bar and a second line of programme text, and the rows
shrank to make it all fit. At three metres that reads as noise.

This takes things away rather than shrinking them. Programme cells carry only a
title; the quality sits as a raised suffix after the channel name; the date
moves into the ruler's left stub; the chrome bar is gone. Rows land at 38dp —
roughly where they already were — but each one holds far less. Entering a group
folds the group column away and gives the guide the full width, and Back or
Left brings it back.

Phone and tablet get the same guide in the shape each screen wants: the phone
drags a bottom sheet up over the player, the tablet keeps a fixed 220dp group
column beside the grid with no landing page in between.

## Commits

- **`fix(live): follow the accent colour chosen in Appearance`** — the guide was
  the one screen in the app with its own colour. `MainActivity` only read the
  `accent_color` key, so a profile carrying the older `focus_border_color`
  resolved to null and all 41 highlights in `live/` fell back to hardcoded cyan.
- **`feat(live): move the sports match guide next to the sports channels`** —
  the match guide sat at the top of the group list as "Sports" while the genre
  filter "Sports · Global" sat far below inside the accordion. They now sit
  together, match guide first. Only the channel list carries a count.
- **`feat(live): a calmer, roomier TV guide, …`** — the layout itself, the
  phone and tablet shapes, and the settings that drive them.
- **`feat(live): open the guide directly on phone, and drop the recents chip`** —
  the phone landed on a group page first, but its groups are already in the
  carousel at the top of the sheet. All three form factors now open on the
  guide. "Recently Watched" had a chip in the carousel while the TV group list
  dropped it — same list, two answers; the chip is gone. Search went from a
  labelled chip to a square icon, with the groups following after it.
- **`fix(live): keep the recents chip gone once recents is the selected group`** —
  the rail adds a chip for the selected category when it is missing, so choosing
  Recently Watched put it straight back, highlighted.
- **`Merge origin/main`** — main moved 48 commits while this was open. Five
  conflicts, all in `live/`; the resolutions are in the merge commit message.
  Worth knowing: main's header tightening to 28dp is kept for the old design,
  and main's `live_cat_sports` replaced this branch's `live_quick_sports`
  mapping.

## Settings

Settings → Sources → TV Guide: **New Guide Design**, **Guide Rows** (6–10, TV
only), **Channels from All Providers**. The switch lives in the TV settings and
cannot be reached from a phone, so touch devices always get the new layout —
asking there would have produced a new frame around old cells.

## Notes for review

- `LiveGuideDensity` is the whole measurement spec as plain Kotlin with no
  Compose imports, so it is unit-testable instead of eyeballed on a device. It
  computes against a 540dp canvas, so "Guide Rows" lands as intended on
  1080p-class screens and is approximate elsewhere.
- Two pre-existing problems surfaced while wiring the touch layouts: the touch
  branch hardcoded `compact = true`, so a 1280dp tablet ran the phone's stacked
  layout and the video filled the content area; and `TouchCategoryRail` had been
  sitting finished and uncalled. Both are fixed here.
- Rows with children ("All Channels", the countries) now expand on click instead
  of entering the group. Previously they did both, and the guide covered the
  column before you could reach the children. This applies to the old design
  too, as it is a fix rather than a variant.

## Verified

Everything below was measured on the merge commit, so after main was folded in.

- 1434 unit tests green (main brought 91 new ones);
  `compileSideloadDebugAndroidTestKotlin` green.
- TV (960×540dp), by `uiautomator` dump: no "Recently Watched" in the group
  list; clicking "All Channels" expands instead of entering; both sports rows in
  the right order; entering a group folds the column away and Back brings it
  back without leaving the app; the match guide carries real fixtures; the ruler
  and its date stub survived main's header change.
- Tablet (1280×800dp): opens on the guide with the fixed column, no recents,
  both sports rows, one Back leaves TV.
- Phone (411×914dp): opens on the guide; carousel is search icon · All Channels
  233 · Sports · Nz 35 · Sydney …; no recents chip, including when Recently
  Watched is the selected category — both after picking it and after a resumed
  session.
- Search, the group drawer and TV focus are untouched since they were measured
  earlier: `git diff` across the later commits does not reach them.

`EpgRefreshRequestsTest.archiveWaitsForShortGuideWithoutStartingConcurrentRequests`
is flaky and untouched by this work — it leaks a coroutine from an earlier test
and has failed then passed on re-run all week. If CI comes back red on it, that
is not this branch.


GAPP99:
Screenshots of the new guide, taken from this branch on emulators with the EPG fully loaded — every programme cell carries a real title, and the header shows time, time remaining, genre, length and description.

## TV — 1920×1080 @ 320dpi (960×540 dp)

Entering the guide, with the group column in place:



Inside a group, where the column folds away and the guide takes the full width. Back or Left brings it back:



## Tablet — 2560×1600 @ 320dpi (1280×800 dp)

Variant T1: a fixed 220dp group column beside the grid, and no landing page in between — the tablet has the room for both at once.



## Phone — 1080×2400 @ 420dpi (411×914 dp)

The guide is a bottom sheet that drags up over the player, with the group carousel in the sheet header: a search icon, then the groups. The phone used to land on a separate group page first; it now opens here, like TV and tablet.



Test playlist: 233 channels from i.mjh.nz.

ProdigyV21:
Thanks for the work you’ve put into this. I do feel the UI changes go too far. We’ve spent a lot of time refining ARVIO’s current design, and this changes its look and behaviour quite significantly. Personally, I don’t see the overall redesign as an upgrade, although that’s subjective.

There are useful improvements here that I’d like to keep:

Making the guide consistently follow the selected accent colour.

Placing the sports match guide alongside the sports channels.

Fixing expandable groups so their children stay accessible.

Fixing tablet detection so tablets don’t use the phone layout.

Allowing users to adjust the number of guide rows.

Could we split those improvements into a smaller PR that preserves the current design? I’d prefer to build on what we have with focused improvements. I also wouldn’t want the new layout forced on mobile and tablet while only TV gets a choice.

GAPP99:
Superseded by #723, which carries exactly the five improvements listed above on top of the current design and nothing else. Thanks for the clear steer.

PR #723, opened 2026-09-18
Live guide: the five focused improvements from #716

GAPP99:
The five improvements you listed on #716, on top of the current design. Nothing else from that branch is here: no redesign, no touch layouts, no landing-page changes. Mobile and tablet render exactly as before except for point 4.

One commit per point, in your order.

### 1. Making the guide consistently follow the selected accent colour

Every highlight in the live guide was `LiveColors.Accent`, a constant, so the guide stayed turquoise while the rest of the app followed the setting. `liveAccent()` reads the same `LocalAccentColorOverride` the player uses and falls back to white; it replaces the constant one-for-one at 54 sites. One read is hoisted out of the now-line `Canvas` lambda, which cannot read a CompositionLocal.

**Worth knowing:** users who never picked an accent see the guide's highlights go from the old turquoise to white — which is what Settings already shows as their accent, and what the player already does. If you would rather keep turquoise as the fallback, it is the one `?: Color.White` in `liveAccent()`.

### 2. Placing the sports match guide alongside the sports channels

The match guide row now sits inside the All Channels accordion immediately before the "Sports · Global" channel list, instead of at the top of the group list. Only the channel list carries a count, which is what tells the two apart. A playlist with no sports channels gets no guide row. On phone the landing page keeps its own Sports tile, so nothing changes there.

### 3. Fixing expandable groups so their children stay accessible

Clicking All Channels or a country both expanded the row and entered the group, and entering closes the drawer — so the children were covered before they could be reached. Those two rows now select their category without closing the drawer (`onSelectKeepOpen`, default `= onSelect`, so every other call site is unchanged). Leaf rows behave as before.

### 4. Fixing tablet detection so tablets don't use the phone layout

One line. The touch branch hardcoded `compact = true` for the mini player, so a 1280×800dp tablet ran the phone's stacked layout with the video alone at 1232×574dp. `liveTvMiniPlayerLayout()` already answers `STANDARD` for tablets (covered by `LiveTvResponsiveLayoutTest`); it is now asked.

### 5. Allowing users to adjust the number of guide rows

Settings → Interface → **Guide Rows**: Auto keeps today's fixed 32dp row; 6–10 divides the guide body into that many rows. `EpgGrid` already measures its own height, so the row height is body height over row count, clamped to stay readable and under the 60dp at which cells begin to show descriptions — the setting changes density, not what a cell shows. TV only; touch layouts keep their finger-sized rows. Stored per profile like the other TV settings.

### Verified

- 1432 unit tests green; androidTest compiles.
- TV emulator (960×540dp): All Channels expands with the drawer staying open and the match guide row directly above Sports · Global; a leaf group still closes the drawer; with Guide Rows = 6, six rows are shown at 39.5dp each; with the accent set to Red, a pixel diff against the default shows the changed pixels at hue 0.
- The setting itself, on screen: Settings → Interface, bottom row, "Guide Rows / Channel rows shown in the TV guide / AUTO". Clicking cycles AUTO → 6 → 7 → 8 → 9 → 10 → AUTO. Set to 6 through the UI, then a cold restart into the guide shows six rows at 39.5dp each.
- Tablet: verified by inspection and the existing `STANDARD` unit test. I could not drive the tablet emulator's landing page from adb to re-measure on this branch; the same expression was measured working on the previous branch.

Strings are English-only, matching the neighbouring `trailer_delay` rows.


GAPP99:
One decision worth your call before this goes in.

`liveAccent()` falls back to white when no accent has been picked, matching the player and what Settings already displays — Accent Color reads "White" by default. The constant it replaces was turquoise, so anyone who never chose an accent will see the guide's highlights change from turquoise to white.

If you would rather the guide keep turquoise as its fallback, it is one line in `LiveTokens.kt`. I did not want to make that call for you.

GAPP99:
Rebased onto current `main` — the branch had gone stale and was conflicting. Three things changed in the process, all worth a look:

**One of the five is gone.** `main` already fixes the tablet mini player: it now passes `compact = miniPlayerLayout != LiveTvMiniPlayerLayout.STANDARD` at that call site, which is exactly what my commit did. The commit became empty and was dropped, so this is four commits now.

**The Guide Rows setting moved from index 45 to 47.** `45` is `setUseForcedSubtitles` upstream now, and `46` is the anime ordering style, so the two collided in the `when`.

**`withSportsDestination()` is reconciled with the sports addon sources.** `main` grew `SportsAddonMatchingTest.sportsDestinationExistsWithoutAnyPlaylistCategories`, which requires the hub to exist even when the playlist has no categories at all — addon sport does not depend on playlist channels, and my version had removed that. It now prefers to sit inside "All Channels" right before `g-sports` when that channel list exists, and otherwise stays a top-level row exactly where `main` put it. Both that test and the ones here pass.

1592 unit tests green locally (`:app:testSideloadDebugUnitTest`). Build Check is sitting at `action_required` and needs an approval from you to run — it has never actually run on this branch.

The question from my earlier comment still stands: `liveAccent()` falls back to white, the constant it replaced was turquoise. Happy to change it to whichever you prefer.

PR #728, opened 2026-09-19
fix(settings): open Trakt activation page with the code prefilled on phones

ReichiMD:
## Summary

On touch devices the Trakt activation dialog only offered "Cancel". To connect
Trakt you had to memorise the eight character code, leave the app, find the
activation page yourself and type the code in by hand.

The dialog already supports an "open auth page" action - `TraktActivationModal`
draws the button whenever `onOpenUrl` is passed and the device is a touch
device. Only the Plex/Jellyfin call site passed it. This wires it up for Trakt
as well, with the code appended to the URL.

## Approach

The Plex/Jellyfin call site is the blueprint. The URL is built the same way
`CloudPairModal` builds the ARVIO account pairing link, just with the code as a
path segment instead of a query parameter.

Measured against the URL the app actually opens:

    curl -sI https://auth.trakt.tv/activate/TESTCODE
    -> HTTP/2 302
    -> location: /signin?callbackURL=%2Factivate%3Fuser_code%3DTESTCODE

So Trakt reads the code out of the path, and a user who is not signed in keeps
the code through the sign-in detour. The host is never hardcoded - whatever
`verification_url` the device-code response carries is used, with the code
appended.

## Scope

- New string `settings_open_trakt_page` in `values` and `values-de`. The
  existing `settings_open_auth_page` stays untouched and remains the default
  label, so Plex and Jellyfin are unaffected.
- Because the code now travels in the URL, the separate "copy code" button is
  redundant for Trakt and is hidden behind a new `showCopyCode` flag. Plex and
  Jellyfin keep it - their URL does not carry the code, so they still need it.
- `traktActivationUrl` is a small pure function next to `openExternalUrl`,
  following the existing `settingsIptvRefreshPolicy` pattern so it can be unit
  tested. It covers a verification URL that already ends in a slash, a missing
  code, and a missing URL.
- The TV layout is untouched. Both phone buttons have always been behind
  `isMobile`, so the QR code path is unchanged.
- SIMKL has the same gap at its call site, but whether `simkl.com/pin/<code>`
  accepts a code in the path could not be verified, so it is deliberately left
  out of this PR.

## Testing

- `testSideloadDebugUnitTest` green: 1446 tests, 0 failures, including five new
  cases for the URL builder.
- Verified on a phone with a sideload debug build: a single button, the browser
  opens with the code already filled in, and the app flips to "Connected" on its
  own without tapping back into the app.
- Re-checked on a TCL C7K (Android TV): the QR code dialog is unchanged.

Screenshot of the new dialog on a phone:
Screenshot_20260919-201225

---


Himanth-reddy:
Can you implement the same thing for simkl?

ReichiMD:
Yes, I wanted to do that too. For all providers as well. If I remember correctly, it's not ideal in the TV view either. I think the login code isn't included with the QR code there. I wanted to take another look at that.

PR #730, opened 2026-09-19
Add an IPTV VOD browser next to TV in the top bar

test01203:
## Summary

Today IPTV VOD is used only behind the scenes: when a TMDB title is opened, the provider's catalogue is searched and a matching copy is added as a source. Users who add an IPTV playlist can't browse what the provider actually offers. This PR adds a **VOD** entry to the top bar, right after TV, that browses the provider's own movie and series catalogue.

### What's new

- **VOD top-bar entry** (`SidebarItem.VOD`) next to TV, plus a matching mobile bottom-bar tab. The top-bar index helpers are derived from `SidebarItem.entries`, so focus and index handling picks up the new item automatically. Every screen's top-bar `when` handles it.
- **`IptvVodScreen`** has:
  - Movies / Series tabs.
  - The provider's categories with title counts, plus "All" and "Recently added". Empty categories are hidden.
  - Search, refresh, and a poster grid with rating and year.
  - TV (D-pad) and touch layouts. It reuses the Library screen's `OledControl` and top-bar pattern.
- **Opening a title:**
  1. If the provider tags a TMDB id, ARVIO's normal Details page opens. The existing VOD matching then lists the IPTV copy as a source next to the addons.
  2. Otherwise the cleaned name (provider prefixes, quality tags and trailing year stripped) plus the year is matched on TMDB.
  3. Movies that still can't be matched play directly from the provider through `Player` with `streamUrl`.
- **Data:** new `IptvRepository.getVodBrowseCatalog()` / `hasVodBrowseSource()`. They reuse the existing Xtream memory and disk caches, so nothing extra is downloaded. `XtreamVodStream` / `XtreamSeriesItem` now also keep `stream_icon` / `cover`, `rating`, and `added` / `last_modified`. The disk cache files are versioned (`vod_v2_`, `series_v2_`) so existing caches, which were written without artwork, are refetched once.
- **Player:** scrobbling and watch history are skipped for `mediaId <= 0`. The direct-play fallback uses a negative synthetic id, so an unmatched IPTV movie never lands in Continue Watching or Trakt with a bogus id.
- English and Hebrew strings.

### Known limitations / follow-ups

- Xtream Codes playlists only, and only the first active one, the same as `getVodCategories()`. Plain M3U VOD entries and Stalker portals aren't browsable yet; the screen points users to IPTV settings.
- Series without a TMDB id or TMDB match can't be opened yet. A provider-side episode picker via `get_series_info` would be the next step.

## Test plan

- [x] `./gradlew :app:compileSideloadDebugKotlin`
- [x] New `CleanIptvTitleTest` (IPTV name cleanup used for TMDB matching) and `OledLibraryLayoutTest` pass.
- [ ] The full `*Iptv*` unit-test run was not completed locally. It was stopped after 20+ minutes with no result, so it's unverified. Please run the full suite in CI.
- [ ] Manual test on TV and mobile with an Xtream playlist:
  - VOD appears next to TV.
  - Switch Movies / Series and categories, and search.
  - Open a TMDB-tagged title, which should go to Details with the IPTV source listed.
  - Open an untagged movie, which should match on TMDB or play directly.
  - Check Continue Watching isn't polluted by direct plays.


Himanth-reddy:
I think this would fit ARVIO better if we implement the VOD browsing within the existing Library instead of adding a separate VOD section next to TV.

I’d keep the TV section unchanged and use it purely for live IPTV. The VOD catalogue from IPTV can be exposed as another source/filter inside Library, since Movies, Series etc. already conceptually belong there.

So I’d suggest keeping the underlying IPTV VOD catalogue/caching/matching work from this PR, but moving the browsing UI into Library rather than introducing a separate VOD destination.

ProdigyV21:
Like Himanth said, adding something in topbar is not in Arvio's vission. Great Pr tho, we can use this. 
Instead add a tab in Library page like it does for homeservers would fit this better. Make sure to fully test it on mobile and tv. Dm me about this and I can give you premium so you can also test for webapp. 

Thanks!

PR #734, opened 2026-09-20
feat(iptv): pick which Stalker catalog categories a lookup searches

ReichiMD:
## Summary

A Stalker portal typically splits its movie and series catalog by language,
one category per language, and a lookup searches all of them. This adds a
per-portal choice of which ones to search, in the settings page that already
manages that portal's live TV groups.

It matters most for series. `findStalkerEpisodeVodSources` binds at most six
shows - two when no `tmdb_id` identified them - in the order the portal sent
them, so a portal listing "Breaking Bad" once per language can spend both
places on languages the user does not speak, and the ones they could watch
are never requested. Filtering before the matcher leaves only the chosen
categories for that cap to spend its places on.

## What changes

- `StalkerApi.getVodCategories()` / `getSeriesCategories()` - the catalog
  counterpart of the `itv&action=get_genres` call live TV already makes.
  Success is measured on the parsed payload: builds without catalog support
  answer with an HTML page or a bare `{"js":""}` under a plain HTTP 200, and
  both come back as null rather than as "this portal has no categories".
- **Both are fetched by the channel load, inside the session that fetches
  the channels, and stored per portal on disk.** Asking when the settings
  screen opens is the one moment they are least likely to arrive: a portal
  closes an idle keep-alive connection after seconds - ten, flat, on the
  one I can measure - the pooled socket is reused anyway, and the request
  is lost with no answer and no retry, which the screen then showed as
  "this portal reports no categories". Fetching them where the live TV
  group names already come from removes that whole class of failure: the
  settings screen reads the device and asks the portal nothing.
- The screen separates "not fetched yet" from "this portal has none".
  Only the second says the portal has none.
- Two profile-scoped preferences, `iptv_hidden_vod_categories` and
  `iptv_hidden_series_categories`, keyed `portalId|categoryId`. Category
  **id**, not name: providers rename their groups, and a selection stored by
  name would quietly fall apart.
- A three-button bar - Live TV, Movies, Series - above the categories page of
  a Stalker portal. Live TV is selected on open and is the page exactly as it
  was, reordering included. Non-Stalker sources get no bar and keep every
  focus index they had.
- The sieve runs on the portal's answer before the matcher, in both the movie
  and the episode path.

## Deliberate non-filters

- **An empty selection searches everything.** A portal nobody configured
  behaves exactly as before, and its source cache key is unchanged, so
  nothing cached by an earlier version is invalidated.
- **An entry without a `category_id` is kept.** The portal not saying where
  something belongs is not the user rejecting it.
- Hiding every category does leave nothing to search - the same thing hiding
  every live TV group already does.

## Not in this PR

- The selection is not carried by the cloud backup, so a second device is
  configured separately. That is a change to `CloudSyncRepository` and a
  separate piece of work; happy to follow up with it.
- Xtream is untouched, although `getVodCategories`/`getSeriesCategories`
  already exist there unused - I cannot test Xtream.
- No category browsing UI. Movies and series stay TMDB tiles; this only
  narrows where a lookup looks.

## Worth knowing before you pick categories

Portals routinely file the same title's quality variants under different
categories - on mine, "Coyote vs. ACME (2026)" sits in `848 DE KINO
HIGHLIGHTS 2026` and "Coyote vs. ACME (2026) 4K" in `849 DE KINO HIGHLIGHTS
2026 4K`. So this narrows by quality as much as by language. That is the
setting doing its job, but it is worth saying out loud.

## Testing

- 1479 unit tests green, 0 failures. 32 new ones cover the two fetches
  (including the HTML and `{"js":""}` answers and the `*` pseudo entry), the
  sieve (empty selection, missing category id, unknown id, cross-portal
  isolation, all hidden), the store (round trip across instances, per-portal
  isolation, a portal that did not answer, one of the two calls failing, a
  re-pointed portal), the cache key, and the focus arithmetic of the bar.
- Built and verified on a TCL C7K (Android TV) and on a phone: the movie and
  series lists load and persist across restarts, bulk hide/show works, a
  deselected group drops out of the source list of a played title, an
  untouched portal finds everything as before, and the live TV tab still
  hides and reorders groups exactly as it did.

Screenshot_20260920-205928

Screenshot_20260920-205735


ProdigyV21:
Thanks for the contribution! Merged after resolving the current-main import conflict and adding a small follow-up: missing movie/series categories now load through a fresh portal session even with Live TV disabled, partial failures remain retryable instead of appearing as empty catalogs, and late responses cannot overwrite another selected tab. Cached lists are retained and reused. Added seven regression tests; both Play/Sideload Kotlin compiles and all 227 selected local tests passed, as did the full GitHub build/unit-test check. Original commits and PR authorship are preserved. Category selections remain device-local as scoped in this PR; I have not independently tested a live portal on a TV.

PR #738, opened 2026-09-21
feat: implement MDBList OAuth 2.0 Device Code flow

Himanth-reddy:
## Summary
Implements the standard OAuth 2.0 Device Code flow (RFC 8628) for MDBList across Android TV and mobile platforms, bringing it into parity with the Trakt and Simkl device linking experiences while removing legacy manual API key entry.

## Changes
- **API & Networking** (`MdbListApi.kt`):
  - Added `oauth/device-authorization/` endpoint for requesting device code, user code, and verification URI.
  - Added `oauth/token/` endpoint for polling token exchange.
  - Added authorization models (`MdbDeviceAuthorizationResponse`, `MdbDeviceTokenRequest`, `MdbTokenResponse`).
  - Added `@Header("Authorization")` support to all MDBList API endpoints.
- **Persistence & Store** (`SyncProviderStore.kt`):
  - Added per-profile DataStore storage for `mdblist_access_token`, `mdblist_refresh_token`, and `mdblist_token_expires_at`.
  - Added backward-compatible fallback for existing API keys and updated provider connection detection.
- **Repository & Polling** (`MdbListRepository.kt`, `SettingsViewModel.kt`):
  - Added device code initialization and polling routines with standard OAuth error handling (`authorization_pending`, `slow_down`, `expired_token`, `access_denied`).
  - Added clean cancellation on dismiss or disconnect.
- **TV & Mobile UI** (`SettingsScreen.kt`):
  - Wired MDBList into `TraktActivationModal` dialog with QR code and external browser launch.
  - Formatted activation instruction URL to display cleanly up to `/device` (`https://mdblist.com/oauth/device`), omitting query parameters while keeping full URL for QR code and browser buttons.
  - Cleaned up background `AccountRow` / `TrackingServiceRow` to avoid inline text expansion and spinners while the modal dialog is open in the foreground.
  - Removed manual API key dialog and input fields.
- **Secrets & Configuration** (`Constants.kt`, `secrets.defaults.properties`):
  - Added `MDBLIST_CLIENT_ID` configuration key.

## Verification
- Built debug sideload build: `./gradlew :app:assembleSideloadDebug -PincludeX86Abis=true` — Successful.
- Tested on Android TV (`emulator-5556`) and Android Phone (`emulator-5554`).
- Updated AST knowledge graph via `graphify update .`.


ProdigyV21:
Thanks for working on this, Himanth. QR-code login would make MDBList much easier to connect on TV. The rebase has resolved the earlier merge conflicts, but let's keep this open until the authentication lifecycle is complete.

These are the remaining issues I found:

1. **Token renewal:** The refresh token and expiry are stored, but no request path uses them to renew the access token. Add serialized renewal, preserve the refresh token when a response omits it, and handle renewal failures without crashing or claiming success.
2. **Disconnect:** Clearing only OAuth tokens leaves an existing API key active. Clear both credential types. Cloud disconnect/import must also clear stale OAuth credentials, not just `mdblist_api_key`.
3. **Cloud and web compatibility:** OAuth access tokens are currently exported as `mdbListApiKey`, without the refresh token or expiry. Keep API keys and OAuth credentials distinct, sync the complete credential, and update Android, the backend's timestamp-based credential merge, and the webapp together. The web proxy currently sends only `?apikey=`, so it does not support the new bearer-token login.
4. **Client ID and fallback:** We don't currently have a confirmed registered MDBList client ID for ARVIO. Please confirm the registration/configuration needed for release builds. Keep the existing API-key login available until OAuth is configured and verified, otherwise removing that dialog leaves new users unable to connect.
5. **Request authentication:** The repository currently sends the same credential as both `Authorization: Bearer ...` and `?apikey=...`. Send OAuth tokens only in the authorization header and legacy API keys only as API keys.
6. **Device polling:** Parse the OAuth error response once. Handle `slow_down`, `access_denied`, and `expired_token` by their error codes, including HTTP 400 responses. The current code only backs off on 429 and can miss standard device-flow responses.
7. **Profile safety:** Bind login/polling and token renewal to the profile that started them. Cancel on profile switch and prevent an in-flight response from reconnecting a disconnected account or saving credentials into another profile.
8. **Unrelated UI changes:** Please retain Trakt/Simkl's existing working indicators and cancellation behavior. Several rows now hardcode `isWorking = false`; this MDBList change shouldn't regress those integrations.

Please add tests for expiry/renewal, legacy-key disconnect, cloud round-trips and disconnects, polling errors, and profile switches. Then verify actual device-code login and cross-device/web behavior with the registered client ID. A successful initial login alone doesn't cover these cases.

I stopped my local fix attempt and discarded those edits. Nothing has been pushed or merged from that attempt; your PR and commits remain unchanged.

PR #739, opened 2026-09-21
Refresh cloud settings on Home return without polling

ProdigyV21:
## Summary
- Preserve Android Home resume cloud refresh; use a monotonic 30-second cooldown and prevent overlapping jobs.
- Refresh web data quietly when returning Home or returning to the browser, sharing a persistent 30-second cooldown with other refreshes.
- Skip automatic web refresh while hidden, playing, or already refreshing. Preserve pending-settings protection and manual refresh.
- No polling, realtime provider, or backend changes.

## Verification
- 623 web tests passed, including Home-return/visibility integration and cooldown checks.
- TypeScript check passed.
- Android Play/Sideload compilation and web production build in progress locally.
- Real two-device authenticated cloud transfer not tested; this does not promise instant propagation.


ProdigyV21:
Local verification complete: Play and Sideload Kotlin compilation passed; all 623 web tests passed; TypeScript and web production build passed. Existing Android Home resume hooks already cover returning Home; this change bounds them and adds the missing web Home-return trigger. No polling or backend changes. A live authenticated two-device transfer was not tested.

